PR #2, opened 2015-08-09
Added Debug Lot setting

ZackCDLVI:
Makes it easier to test different debug lots from the server. Also is TargaImage was also missing from the solution. Is this still needed? I see it's still in FSO.Files still.


ZackCDLVI:
Fixed. :D

RHY3756547:
TargaImage's functions are never used by the standalone server, which is why it was working correctly without including it. You're right though, it should be there.

PR #4, opened 2015-10-07
Added ui panel thunbmail button

francot514:
Should check why ucp selectedavatar is not working, maybe becuase sim is created by net command


LetsRaceBwoi:
Shouldn't `public UIButton MoodPanelButton;` be `public UIButton MoodPanelButton { get; set; }` like the rest?

RHY3756547:
No, the auto-properties are only for controls to be set by the UIScript. livepanel.uis has no such definitions, it only provides moodpanel positions and the mood overlay images.

You seem to be literally replacing the thumbnail frame with the thumbnail itself, which can't be right. I'll merge, but I'll probably need to change that a bit. Also fix the live panel being totally unfinished. That might be important.

PR #7, opened 2015-10-20
Include FSOMonogame as part of the project.

RHY3756547:
Lets us better debug issues with Monogame, and possibly test out implementing features not found in monogame mainline. Need to change build bot to run the monogame build scripts.


RHY3756547:
Here goes nothing! Servo should build this I think...

PR #10, opened 2016-05-23
Updated Simslib with Sims 1 chunks compatibility

francot514:
Some sims 1 chunks added to be completed

### DIFF
--- a/Other/tools/SimsLib/SimsLib/IFF/ARRY.cs
+++ b/Other/tools/SimsLib/SimsLib/IFF/ARRY.cs
@@ -1,0 +1,102 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.IO;
+
+namespace SimsLib.IFF
+{
+
+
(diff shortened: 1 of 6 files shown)


RHY3756547:
You seem to have reverted SimsLib's "Iff" to the old format (which is a literal super mess). I also already said that SimsLib is outdated, so you should probably add these to FSO.Files instead (more up to date in every way).

francot514:
Youre not considering that sims 1 uses old format compatibility, and FSO.Files is targeted to TSO so is not really best idea

RHY3756547:
FSO.Files and FSO.SimAntics will eventually be cross compatible. SimsLib is only in the codebase because those tools have not been moved off of it. FSO.Files is the evolution of SimsLib, it doesn't make sense to be going backwards.

francot514:
Ok, no problem so "FSO" will be change eventually? to FS1??

RHY3756547:
No, because that wouldn't represent the cross compatability. Probably something like FSX.

PR #35, opened 2016-11-13
Feature/group money objects

ddfczm:
Working paper chase (code) group money object. Unlike Pizza, I did not see any events fired in to tell the controller when to move onto the next state so I have implemented it based on timers. I think this is wrong but need a second opinion. However, it does seem fully functional.

Note, there seems to be a bug with normal height EODs not positioning the UI correctly. Not had chance to look at this.


RHY3756547:
LGTM, might even change pizza object to use the more general EODLobby class. Will look into EOD UI template in a wee bit.

RHY3756547:
It's just a 20px offset I left in by accident. I never really got the chance to test straight type EODs, so I never caught it.

PR #37, opened 2016-11-29
Clothing racks + dressers

ddfczm:
Full implementation of store clothing racks and dressers to allow sims to change outfits. This does not include costume trunks.

Change list:
* Content loader for rack packingslips (purchasable.xml)
* DB table to store purchased outfits (fso_outfits)
* Migration added to give all existing sims their current outfits in the dresser
* VMAvatar now has slots for DynamicSuits and Decor (Version >= 15)
* Added a NetCmd to set suit values
* EODs now supports double tall
* EODs use a bit less hard-coded logic and a bit more of the UIScript
* Added a debug EOD to explore all EOD combinations
* Added a UIRadioButton control for easier radio groups
* Updated UIScript to use UIRadioButton if AddButton is called and the target class has a property of type UIRadioButton. Otherwise it uses UIButton
* Fixed EOD icon not showing
* Rack owner EOD plugin
* Rack EOD plugin
* Dresser EOD plugin
* UICollectionViewer now supports multiple kinds of pagination style
* Updated lotadmit change script to remove tablename (I use a different table name and others might too)
* IDE now shows local labels for semi globals and globals
* IDE now shows subroutine arguments in the description
* Added a VBox and HBox container type (used in debug EOD). It auto positions the controls horizontally or vertically. (I found this quicker than trying to manually place pixels). Might be useful for custom UI in the future but not for anything built in.

Constraints:
* You are allowed 20 items in a rack
* You are allowed 5 items per category in the dresser

Regression:
* All existing EODs have been tested
* Online job UI to ensure not broken
* CAS (because UICollectionViewer changed)
* All behaviors of racks and dressers

Known issues
* Some of the buttons in the dresser have masking / alpha issues
* Accessory / decor racks move the sim vertically off screen when browsing
* Cancelling a rack / dresser EOD won't exit the EOD but pressing the close button will
* My local version failed to load my lot, should just be because we had an overlap in marshal version. Having made a new lot all is fine, but something to be aware of.
* Race condition where sometimes the avatar changes into the defaultdaywear outfit rather than the chosen outfit in dressers. Think this is a temp mixup but cant prove it.

Unknowns:
* I am not using any kind of price multiplier for hard mode
* Not sure if racks gave bulk discounts. Nothing implemented for this currently.


TSOClient/FSO.Server.Database/DatabaseScripts/changes/0005_lotadmit_primarykey.sql
```diff
@@ -1,3 +1,3 @@
-﻿ALTER TABLE `fso`.`fso_lot_admit`
```
RHY3756547: Must have totally blanked out when writing this. Will definitely avoid in future.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODRackOwnerPlugin.cs
```diff
@@ -0,0 +1,127 @@
...
+            if (outfit == null) { return; }
+
+            var VM = client.vm;
+
+            //TODO: Do stores get bulk discounts on clothes?
+            VM.GlobalLink.PerformTransaction(VM, false, client.Avatar.PersistID, uint.MaxValue, (int)outfit.Price,
+                        
+                (bool success, int transferAmount, uint uid1, uint budget1, uint uid2, uint budget2) =>
+                {
+                    //TODO: Make this part of global link
```
RHY3756547: Yeah, will probably just move this in and include a "thread id" parameter for the global link call. I don't see any cases where we wouldn't want to update the client's view of budget after a transaction.

TSOClient/tso.simantics/packages.config
```diff
@@ -1,4 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="MonoGame.Framework.WindowsGL" version="3.4.0.459" targetFramework="net4" />
+  <!--Although we don't do any network stuff, this gives us access to IoBuffer for serialization-->
+  <package id="Mina" version="2.0.11" targetFramework="net4" />
```
RHY3756547: Not sure about this. What is using IoBuffer right now?

ddfczm: Serialization of outfits, it just needs IoBuffer. we could always rip that part out into common.

PR #38, opened 2016-12-04
House button in SAS

ddfczm:
Ability to connect directly to a lot from SAS. House thumbnails also cache locally and are used for SAS.


RHY3756547:
LGTM. Should make sure this ties into terrain - and our own lot thumbnail should probably update as soon as we render it.

PR #39, opened 2016-12-10
EOD: Scoreboard

ddfczm:
Continuing through the list of EODs. Still trying to make each one easier to build by adding archetypes and utilities.

This implements the scoreboard + some utilities for similar EOD types.


RHY3756547:
LGTM, might be a while til I can refactor the old ones to use the new classes (lobby and single player).

Just a heads up, with the new scheduler live debug with volcanic might be even less reliable. I'll probably pass over all volcanic and sandbox mode stuff when I have more time.

PR #41, opened 2016-12-19
Fix for Mac Support

Bengrs:
This commit add a Game Locator that allows Mac user to store TSO Game
files in the Documents folder. MonoGame for Mac as been replaced with a
Xamarin based version. Warning: FSO.Client won’t run without a
kickstart program.

Indentation problem fixed.
Checking for MacOS using Directory.Exists("/Users") is not perfect but no other simple solution yet.
Binary included are modified FSOMonogame.


RHY3756547:
LGTM.

PR #44, opened 2017-01-05
Add MRT commands for legacy hardware users

sinkillerj:
This is intended as a workaround for users with the multiple render targets issue discussed here: http://forum.freeso.org/threads/call-for-testing-client-with-dx9-support-for-intel-gma.1079/page-2#post-23602

This requires testing prior to being merged as I currently do not have an environment setup to compile the changes.

### DIFF
--- a/TSOClient/tso.client/Program.cs
+++ b/TSOClient/tso.client/Program.cs
@@ -48,6 +48,7 @@ namespace FSO.Client
                 gameLocator = new WindowsLocator();
 
             bool useDX = false;
+            bool useMRT = true;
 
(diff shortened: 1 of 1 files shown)


RHY3756547:
I'm not sure if this will work out of the box, last time I tested MRT was in conjunction with our SoftwareDepth implementation. Will test.

sinkillerj:
Ill be attempting to test this myself in a day or so as well, just have to get visual studio downloaded.

sinkillerj:
Looks like this PR is a bust for the reasons stated in the main topic, so closing for now.

Issue #49, opened 2017-03-01
Neighborhood View: Client responds to 4K screen resolution only in the top left quarter


Itja:
Windows 8.1, screen resolution **3840x2160** (4K), Version beta/update-16:
On the neighborhood view, where moving the mouse curser normally results in a squares on the ground to be drawn, this only happens in the top left quarter of the screen. It also does not react to left clicking in the area where the squares are not drawn. Right clicking in the bottom right area results in ultra-fast screen movement towards the bottom right (which is probably caused by the client taking the mouse distance from the presumed "center", which is here in the top left quarter, to determine the scroll speed).

Resizing the window back to 1080p resolution is a workaround to this issue. This issue does not arise inside of lots (there, 4K works fine).

Itja:
I found probably the cause of this. It seems to be caused by the ScreenWidth and ScreenHeight only used once for the MouseListeners on initializing the Screen. Further evidence is that the problem does not occur if I start the game in 4K resolution - then, I can use the whole city screen normally without issues. I'll try to create a patch for this and submit a pull request when I'm done - only if you're not actively working on this part of the code at the moment, of course.

RHY3756547:
That sounds about correct. I'm not working on anything this week due to university coursework, so feel free to make changes.

PR #50, opened 2017-03-03
City screen: resize mouse hit area on window resize

Itja:
This fixes Issue #49.


RHY3756547:
LGTM. This stuff needs to be cleaned up in general too, I think the tooltip that shows lot prices and names got completely broken at some point trying to keep it out of the lot view gameplay.

PR #51, opened 2017-04-09
Fully Functional Slot Machines (entertainment)

TheArchitectFreeSO:
These changes make the three slot machine objects fully functional.

Since FreeSO does not allow objects to go into debt, each slot machine has a balance min-max range of $2500-7500, $12500-17500, and $25000-32500 for the $1, $5, and $10 slot machines, respectively.  These minimums have been calculated to cover the maximum payout possible from a "jackpot" spin, which is always 500x the player's bet (see below).

Players can choose to "Play $% Slots" where % is "10" for the "'Jack of Hearts' Slot Machine", "5" for the "'Gypsy Queen' Slot Machine" and "1" for the "Viva PGT[sic] Home Casino".  Each machine allows the player to bet one to five "chips" of their respective increments (1, 5, and 10) before pulling the slot machine handle and spinning the wheels.  When the player chooses their bet amount and pulls the handle, the server checks to make sure the sim has the money, debits their budget and pays the object's budget, spins the wheels, and checks the results.  If the player wins, the server debits the object's budget and pays the player's budget. After every single game, the server checks the machine's balance to make sure it is within the appropriate range before starting a new game.  If the object's balance if out of range, the object enters the "off" state, telling the player that the machine is closed for maintenance.

All newly purchased slot machines begin with a balance of $0. Owners MUST deposit a balance from the "manage" interaction by clicking the "cash out" button and choosing "Deposit" BEFORE players will be able to play.  If a machine does not have enough stocked simoleons, has too many stocked simoleons, or is powered "off" (a bubble button found in the "manage" interaction's UI Panel), then the interaction to "play" will not be found in the pie menu when a user clicks the object.

Finally, owners can set the "odds" of a machine. This has proven to be a very convoluted concept, but with the help of TSOMatt and his impressive TSO memory, I have decided that the number displayed should be considered the machine's "Return to Player" or RTP. 

RTP is a gambling industry standard calculation that adds together all of the probabilities of winning combinations of slot stops multiplied by their payout rate in order to determine the "fairness" of a slot machine.  Usually when manipulating RTP in the favor of the house (less than 100%) the payout amounts change, leaving the odds of any winning combination to remain the same.

However, due to the cryptic use of the word "Odds" that Maxis chose to label this slider, as well as TSOMatt's recollection that all payout schemes remained ubiquitious across each machine despite the "Odds" setting, I decided this slider should directly affect the probability of each winning combination. Therefore, when the owner manipulates the "Odds" slider, he is literally manipulating the odds of a player winning as it is programmed to directly affect the number of winning stops and blank stops available in any one game.  However gambling experts and enthusiasts should still regard this slider number as representing RTP.

Known Issues:
•Certain sound effects do not play despite being called from the slot machine object "play" tree: "slot_pull_handle" , "slot_win" , "slot_lose" and "slot_tantrum_vox"
•The "?" buttons on the UI in "Play" and "Manage" do nothing
•Depositing and Withdrawing in "manage" mode do not play the classic "slot_coin_drop_in" sound effect
•Amounts of simoleons paid or rewarded are not displayed in a box over the sims' heads, as is currently congruous with all other interactions in FreeSO


TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
```diff
@@ -0,0 +1,678 @@
...
+                        machineBalance = -1;
+                        userClient.Send("slots_player_init", new byte[3] { Convert.ToByte(args[1]), Convert.ToByte(args[2]), 0 });
+                        State = VMEODSlotsStates.Off;
+                    }
+                });
+            }
+            else // user is owner
+            {
+                // I'm sure there's a more elegant way, but get the type of machine by the 2nd letter in its name
+                var name = Server.Object.Name;
```
RHY3756547: Love it.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
```diff
@@ -0,0 +1,678 @@
...
+                    Convert.ToByte(tempMachineBalance / 255), Convert.ToByte(args[4]), machineType });
+            }
+        }
+        public override void OnDisconnection(VMEODClient client)
+        {
+            base.OnDisconnection(client);
+        }
+        private void ToggleOnOffHandler(string evt, string OnOffState, VMEODClient client)
+        {
+            onOffState = Int16.Parse(OnOffState);
```
RHY3756547: I know ddfczm does this for the rack, but an invalid value here would _crash the VM_. I would use TryParse here, though just as a precaution I would add a try catch around vm.EODHost.Deliver() in VMNetEODMessageCmd.

RHY3756547: Precaution for developers using the unsafe parse methods, I mean. I would not rely on an arbitrary try catch to save anyone from potentially dangerous crashes while changing state.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
```diff
@@ -0,0 +1,678 @@
...
+            client.SendOBJEvent(new VMEODEvent((short)VMOEDSlotsObjectEvents.ToggleOnOff));
+        }
+        private void NewOddsHandler(string evt, Byte [] args, VMEODClient client)
+        {
+            machineOdds = args[0];
```
RHY3756547: Are these odds within bounds? What do we allow?

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
```diff
@@ -0,0 +1,678 @@
...
+            client.SendOBJEvent(new VMEODEvent((short)VMOEDSlotsObjectEvents.SetNewPayout, new short[1] { machineOdds }));
+        }
+        private void WithdrawHandler(string evt, string amountString, VMEODClient client)
+        {
+            // parse the withdraw amount
+            short withdrawAmount = Int16.Parse(amountString);
```
RHY3756547: Another uncaught parse.

RHY3756547: Also... Is this non-negative? A negative transaction would actually deposit money into the machine...

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
```diff
@@ -0,0 +1,678 @@
...
+                    client.Send("slots_resume_manage", "0");
+                }
+            });
+
+            // client.SendOBJEvent(new VMEODEvent((short)VMOEDSlotsObjectEvents.WithdrawCash)); *Depricated*
+        }
+        private void DepositHandler(string evt, string amountString, VMEODClient client)
+        {
+            // parse the withdraw amount
+            short depositAmount = Int16.Parse(amountString);
```
RHY3756547: Ditto but in reverse. A negative value could withdraw money from the machine. (plus needs machine owner check)

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
```diff
@@ -0,0 +1,678 @@
...
+            }
+            else
+            {
+                State = VMEODSlotsStates.Off;
+                userClient.Send("slots_toggle_offline_message", "");
+            }
+        }
+        private void BetHandler(string evt, string betAmountString, VMEODClient client)
+        {
+            short betAmount = Int16.Parse(betAmountString);
```
RHY3756547: The user can literally steal money from the machine if they give a bad value here, and you haven't bounded what they can bet. If a user submitted a request with a modified client to bet a million simoleons, it would do it.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
```diff
@@ -0,0 +1,678 @@
...
+            // attempt to debit player the bet amount
+            var VM = client.vm;
+            
+            VM.GlobalLink.PerformTransaction(VM, false, client.Avatar.PersistID, Server.Object.PersistID, Convert.ToInt32(betAmount),
+
+            (bool success, int transferAmount, uint uid1, uint budget1, uint uid2, uint budget2) =>
+            {
+                // add bet to the balance in object
+                machineBalance += betAmount;
```
RHY3756547: budget2 contains the machine's true balance. Why not set it to that?

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
```diff
@@ -0,0 +1,678 @@
...
+                    TransferAmount = transferAmount,
+                    UID1 = uid1,
+                    Budget1 = budget1,
+                    UID2 = uid2,
+                    Budget2 = budget2
+                }));
+                if (success)
+                {
+                    // roll the slots rng and get result, send to UI and send to SlotResultHandler() to determine win
+                    SlotResultHandler(RollNewGame(betAmount));
```
RHY3756547: This is running in the background thread (callback from global link). Modifying state from outwith the main loop of the VM could cause a race condition. The VM.SendCommand and userClient send functions are thread safe, everything else is up to you.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
```diff
@@ -0,0 +1,678 @@
...
+                        Success = success,
+                        TransferAmount = transferAmount,
+                        UID1 = uid1,
+                        Budget1 = budget1,
+                        UID2 = uid2,
+                        Budget2 = budget2
+                    }));
+                    if (success)
+                    {
+                        machineBalance -= winnings;
```
RHY3756547: New machine budget is in budget1.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
```diff
@@ -0,0 +1,678 @@
...
+        private void WheelsStoppedHandler(string evt, string payoutString, VMEODClient client)
+        {
+            if (payoutString[0] != '0')
+            {
+                State = VMEODSlotsStates.DisplayingWin;
+                // send win event with random number for message displayed
+                client.Send("slots_display_win", "" + slotsRandom.Next(25, 30) + "%" + payoutString);
+
+                // pay the player
+                short winnings = Int16.Parse(payoutString);
```
RHY3756547: Did you just get the payout value... From the client? Then use it to give someone money? I cannot accept this pull request if something like this is possible.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
RHY3756547: ALSO - Is the person using this actually the owner of the machine? As is, looks like anyone can turn the machine off.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
RHY3756547: ALSO: can this be set by non-owner?

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
RHY3756547: AND! Is the avatar actually using the plugin the owner? If a client sends this command right now, wouldn't they be able to steal money from the machines?

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
```diff
@@ -430,57 +485,39 @@ private void SlotResultHandler(int[] betAndSlotResults)
...
             {
-                State = VMEODSlotsStates.DisplayingWin;
                 // send win event with random number for message displayed
-                client.Send("slots_display_win", "" + slotsRandom.Next(25, 30) + "%" + payoutString);
+                client.Send("slots_display_win", "" + SlotsRandom.Next(25, 30) + "%" + CurrentWinnings);
 
                 // pay the player
-                short winnings = Int16.Parse(payoutString);
-                
                 var VM = client.vm;
```
RHY3756547: Set current winnings to 0 immediately before calling this, and use a copy instead. Since there is a delay between a transaction starting and completing, current winnings will still be greater than 0 for a split second after this handler is run, so the handler could be run again to get a double payout.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODSlotsPlugin.cs
```diff
@@ -183,201 +255,177 @@ public override void OnConnection(VMEODClient client)
...
+                        // dispatch event to tell object that the user has lost
+                        UserClient.SendOBJEvent(new VMEODEvent((short)VMOEDSlotsObjectEvents.ExecuteLoss, new short[1] { betAmount }));
+                        }
+                        // send spins results to UI to simulate gameplay
+                        UserClient.Send("slots_spin", new Byte[] { (byte)SlotsStops[thisGameBetAndResults[1]], (byte)SlotsStops[thisGameBetAndResults[2]],
+                        (byte)SlotsStops[thisGameBetAndResults[3]] });
+                    }
+                    else
+                    {
+                        CurrentWinnings = 0; // forfeit any winnings
```
RHY3756547: Someone could bet money they don't have to set the CurrentWinnings, then try to get the winnings before the game realizes they cannot be paid for the bet.

Store the winnings in a local variable outside of the transaction, then ONLY set CurrentWinnings to that value when the transaction succeeds.

PR #53, opened 2017-04-13
Slot Machines: Tweaks Redux

TheArchitectFreeSO:
•Fixed an issue that prevented non-English clients from playing $5 and
$10 machines
•Increased the maximum balances of all machine types


RHY3756547:
LGTM. Thanks for fixing CAS crash, I always forget about that one.

PR #55, opened 2017-04-14
TrunkEOD: Done

TheArchitectFreeSO:
All 9 costume trunk objects now work. Simple as that.

The only known issue is that the rummage loop animation set to run in the background does not work.


RHY3756547:
LGTM. I'll remove that Console.WriteLine.

PR #56, opened 2017-04-23
Rack Owners can now name their racks & Other UI polishing tweaks

TheArchitectFreeSO:
•Rack owners can name their racks, and the name is saved on the server and loaded by the next user, customer or owner.
•Only 32 characters are allowed for rack names.
•A couple more Parse methods were replaced with TryParse for safety.
•A few additional checks were made to functions to ensure the calling clients belong to the rack owner
•MaxChars was added to Alert Options
•UITextEdit MaxChars functionality was copied into UITextBox
•Doors and Slots EOD now limit the number of characters the users can enter into the alert.


RHY3756547:
Good work, LGTM. We will merge UITextEdit and UITextBox at some point in the future, but changes to clean it up for now are welcome.

PR #57, opened 2017-04-28
New EOD: WarGame

TheArchitectFreeSO:
WarGameEOD now works. I made one .piff change to the WarGame object to add an event where the internal variables counting the players' round wins could be reset. This change was needed so that round wins from unfinished games (from disconnects) or stalemates don't carry over to next new game (cf. pizzamaker "player contributions" internal variable issue).


TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODWarGamePlugin.cs
```diff
@@ -0,0 +1,359 @@
...
+            short playerNum = -1;
+            if (BluePlayerClient.Equals(playerClient))
+                playerNum = (short)VMEODWarGamePlayers.Blue;
+            else if (RedPlayerClient.Equals(playerClient))
+                playerNum = (short)VMEODWarGamePlayers.Red;
+            else
+                return;
+
+            // verify that the piece is a legal selection
+            bool isLegal = Enum.IsDefined(typeof(VMEODWarGamePieceTypes), chosenPieceNum[0]);
```
RHY3756547: Saving that space with the single byte binary message, love it.

RHY3756547:
Very impressive - you're on fire! Not much to comment on here since there's not much attack surface (two basic UI->VM messages), and the UI is looking very clean. LGTM.

PR #59, opened 2017-05-06
EOD: GameComp(onent)DrawACard

TheArchitectFreeSO:
What I thought would one of the simplest EODs turned out to be the most complicated! I was very verbose with my comments for the server plugin because of the complicated user data handling and validation processes.

The .piff file adds the missing connection events required to instantiate the VMEODServer, as well as one custom event that tells the object that no card has been drawn, removing the "View Current Card" interaction.

A maximum number of unique cards was set to 100, but is easily changed. The plugin does not support stocking 0 of any unique card (it must be minimum 1). Up to 99 copies of any unique card can be put into the deck. The plugin allows cards to be blank, and cards can have up to 256 characters.

Any new object copies (which essentially will fail to load any data from the server upon first use) comes with 19 default cards, a default game name, and a default game description.

I'm sure there are bugs...but I hope there aren't!  The more testing of different types and amounts of input is what this plugin needs, so I feel confident in its deployment.


RHY3756547:
LGTM. Looks like something weird might happen if the data takes too long to load, but I'm not too sure what.

PR #63, opened 2017-05-17
EOD: WarGame bug fix

TheArchitectFreeSO:
Fixed logic error for determining winner.


TheArchitectFreeSO:
EOD: GameCompDrawACard:
The server loads 100 unique cards very quickly.  I'd like to test it
with 300, due to popular demand. Also a config fix with the project file
because I pasted a piff to the wrong directory.

RHY3756547:
LGTM. Missed this as I was super busy 🥇

PR #68, opened 2017-08-23
[API] Add lot categories and descriptions to city.json

AstroPsynapse:
Added lot categories and descriptions to city.json and modified the SQL query to grab the necessary information.


RHY3756547:
The way you've done this isn't looking great for our production server. Dumping and serving out the descriptions for over 2000 lots is going to take a toll on the server. Since a ton of tools use this json and ignore the descriptions, it's probably a good idea to serve them through some other source.

I think you should instead allow specific requests for each lot to get info like description. They should be similar to the path to its thumbnail png, but instead end in .json.

You can also instead make another request that provides more information for only lots that are online. This would save making a ton of requests for each lot.

RHY3756547:
Are you going to consider implementing this, or should I close the pull request and open an issue for someone else to take on?

thertzelle:
dotequals has suggested that I take on this request if astropsynapse isn't going to completed it.

AstroPsynapse:
Apologies for the lack of response. It'd probably be best for someone else with a better understanding to take on this task.

PR #70, opened 2017-09-16
[Bugfix] Read extra unknown byte from TTAT chunk data

gmoben:
- Fixes crash on loading empty neighbourhoods via Simitone (possibly more)

### DIFF
--- a/TSOClient/tso.files/Formats/IFF/Chunks/TTAT.cs
+++ b/TSOClient/tso.files/Formats/IFF/Chunks/TTAT.cs
@@ -45,6 +45,7 @@ namespace FSO.Files.Formats.IFF.Chunks
                         tatts[j] = iop.ReadInt32();
                     }
                     var unknown = iop.ReadInt32();
+                    var unknown2 = iop.ReadInt32();
                     TypeAttributesByGUID[guid] = tatts;
                 }
             }


RHY3756547:
There's something else going on here. My TTATs do not have this unknown int - reading it causes a misalignment, where the second ttat collection reads a guid of "32", then a count of "0"... when that should actually be the attribute count. The remaining ttats are mostly filled with zeroes since they are the  32 elements of the first entry.

Can you post your neighbourhood folder zipped up?

gmoben:
@RHY3756547 https://drive.google.com/open?id=0B6eh2nz1sZIXZHVBZ2VaSE9Nd3M

I've included the default neighborhood (with tutorial) as well as an empty initialized and uninitialized neighboorhood. The alignment is one byte off for my original neighbourhood too and I think I've traveled to expansion lots and saved via the Maxis client. I'm skeptical of Old Town lots since I couldn't travel in Simitone without this patch. 

Let me know if you figure something out, it's been interesting looking at the structure of these files.

gmoben:
If there multiple versions somehow of this chunk type, we could always catch the exception and attempt reading with and without the extra byte as a stopgap.

RHY3756547:
This isn't a normal format - as you've seen the data is obtained through an "IoProxy", which reads int or short size elements from a TS1-specific field encoding: http://simtech.sourceforge.net/tech/misc.html#fields

Reading integers with the loop commented out above the actual reading is the easiest way to get a dump of the data. I'll compare readings from my TTAT and yours to see if there's some condition for that second unknown existing.

Looks like you need to approve the google drive access.

gmoben:
@RHY3756547 yeah niotso and simtech had very helpful documentation. I'm glad to see good write ups on this stuff. I found this out by stepping through the `Read()` and saw it was off by one byte. I fixed the permissions, so you can view them now.

RHY3756547:
I figured it out. The actual TTAT data is using the short bit-widths in the encoding. There are no unknown fields in the format. The old code was working on my save file due to sheer luck.

```csharp
                        tatts[j] = iop.ReadInt16(); //was 32 before
```

It's definitely working now - starting a lot no longer asks if I want the help system on or tells me about vacation island. Will commit fix shortly.

PR #71, opened 2017-09-17
Adding the ability to resize the Property Chat size and font. 

thertzelle:
This PR gives the users the ability to increase or decrease the font/window size within the FreeSO Client. 

This feature adds a keyboard shortcut, similar to the Control + H shortcut, that gives the user the ability within Property Chat to press Control +, or Control - (Control + OemPlus, Control + OemMinus).

As a user of FreeSO since early April, I have seen a common request from beginning players to "founders" whom would love to see this feature added. We're all older now and it would be great to be able to bump up the font size to allow us to read the chat better and better converse with our friends.

If this PR is successful, I'd like to continue to work on this feature by implementing it within the messages and perhaps even give it a graphical button, or a setting that would allow users to keep their preference more permanently. But for now...small steps.

Thank you for your time, and for building such a neat code base.

### DIFF
--- a/TSOClient/FSO.UI/Framework/Font.cs
+++ b/TSOClient/FSO.UI/Framework/Font.cs
@@ -50,6 +50,11 @@ namespace FSO.Client.UI.Framework
                 EntryList.OrderBy(x => Math.Abs(pxSize - x.Size)).FirstOrDefault();
         }
 
+        public FontEntry GetIdeal(int pxSize)
+        {
+            return EntryList.First(x => x.Size == pxSize);
+        }
+
         public void AddSize(int pxSize, SpriteFont font)
         {
             EntryList.Add(new FontEntry {
--- a/TSOClient/FSO.UI/Framework/Font.cs
+++ b/TSOClient/FSO.UI/Framework/Font.cs
@@ -50,11 +50,6 @@ namespace FSO.Client.UI.Framework
                 EntryList.OrderBy(x => Math.Abs(pxSize - x.Size)).FirstOrDefault();
         }
 
-        public FontEntry GetIdeal(int pxSize)
-        {
-            return EntryList.First(x => x.Size == pxSize);
-        }
-
         public void AddSize(int pxSize, SpriteFont font)
         {
             EntryList.Add(new FontEntry {
--- a/TSOClient/tso.client/UI/Panels/UIChatDialog.cs
+++ b/TSOClient/tso.client/UI/Panels/UIChatDialog.cs
@@ -10,12 +10,6 @@ using System.Text;
 
 namespace FSO.Client.UI.Panels
 {
-    public static class SizeFactor
-    {
-        public const int Decrease = -1;
-        public const int Increase = 1;
-    }
-
     public class UIChatDialog : UIDialog
     {
         public event UISendMessageDelegate OnSendMessage;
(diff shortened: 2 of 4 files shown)


TSOClient/FSO.UI/Framework/Font.cs
```diff
@@ -50,6 +50,11 @@ public FontEntry GetNearest(int pxSize)
                 EntryList.OrderBy(x => Math.Abs(pxSize - x.Size)).FirstOrDefault();
         }
 
+        public FontEntry GetIdeal(int pxSize)
```
RHY3756547: Why add&use use this when GetNearest has the same function, assuming pxSize exists within the font?

TSOClient/tso.client/UI/Panels/UIChatDialog.cs
```diff
@@ -10,6 +10,12 @@
 
 namespace FSO.Client.UI.Panels
 {
+    public static class SizeFactor
+    {
+        public const int Decrease = -1;
+        public const int Increase = 1;
+    }
```
RHY3756547: No need to make this an enum - for simple deltas like 1 and -1, it makes more sense to just pass in the delta unnamed.

For future reference, though, enums in C# work as follows
```csharp
    public static enum SizeFactor
     {
         Decrease = -1,
         Increase = 1
     }
```

TSOClient/tso.client/UI/Panels/UIChatDialog.cs
```diff
@@ -114,6 +122,39 @@ public override void Draw(UISpriteBatch batch)
...
+
+            ChatHistorySlider.Position = new Vector2(ChatHistorySlider.Position.X + updatedSize, ChatHistorySlider.Position.Y);
+            ChatHistorySlider.SetSize(ChatHistorySlider.Size.X, ChatHistoryBackground.Size.Y - 26);
+            ChatHistoryScrollUpButton.Position = new Vector2(ChatHistoryScrollUpButton.Position.X + updatedSize, ChatHistoryScrollUpButton.Position.Y);
+            ChatHistoryScrollDownButton.Position = new Vector2(ChatHistoryScrollDownButton.Position.X + updatedSize, ChatHistoryScrollDownButton.Position.Y + updatedSize);
+
+            ChatEntryTextEdit.Position = new Vector2(ChatEntryTextEdit.Position.X, ChatEntryTextEdit.Position.Y + updatedSize);
+            ChatEntryTextEdit.SetSize(ChatEntryTextEdit.Size.X, ChatEntryTextEdit.Size.Y);
+            ChatEntryBackground.Position = new Vector2(ChatEntryBackground.Position.X, ChatEntryBackground.Position.Y + updatedSize);
+            ChatEntryBackground.SetSize(ChatEntryBackground.Size.X + updatedSize, ChatEntryBackground.Size.Y);
```
RHY3756547: Resizing the window with the text size is a good idea. In future, it might be useful to be able to resize this window by dragging on the edges. I believe it was the only window you could do that on in tso.

TSOClient/FSO.UI/Framework/Font.cs
thertzelle: I will re-review GetNearest, I was having unexpected behavior when using it. It seemed like based on the px it would only retrieve two of the 4(?) available sizes.

TSOClient/tso.client/UI/Panels/UIChatDialog.cs
thertzelle: Some people like it for documentation purpose, I figured I was being clever just in-case you wanted it to be self-documenting. I decided to not use the enum because I didn't want to cast (int)factor later on, I'll reduce this to just the delta.

TSOClient/tso.client/UI/Panels/UIChatDialog.cs
thertzelle: Yeah in the future I'd like to make this even smarter, I was unsure if this project would even accept pull requests, so I put in a version that is kinda the best of both worlds for the initial version, to see what the reception from developers to users would be.

thertzelle: I would love to implement the re-size window feature in the near future.

TSOClient/FSO.UI/Framework/Font.cs
RHY3756547: If there's a problem with that, it might be worth finding out what's wrong there, since it's used for all other font rendering.

thertzelle: Its working as intended now, when I was new to the code that was one of the first things I tried to tackle and something wasn't mixing well before, probably just me getting my feet wet with the code, my update of this PR removes a lot of excess code as the function (GetNearest) works as intended.

thertzelle:
Requested Changes have been modified. Most of the changes to Font and FontEntity have been reversed in favor of using GetNearest. Removed the increase/decreasing static/class/enum.

Thanks for the feedback.

PR #72, opened 2017-09-17
Better User Experience for Login Workflow.

thertzelle:
This PR adjusts the workflow for an incorrect password, giving a better User Experience for users that mistype their passwords, further more, it helps users return to re-enter their password and allows adjustments for incorrectly typed usernames with a Shift + Tab functionality. This PR also adjusts the password workflow such that it no longer spams the login services with incorrect credentials when users press enter again after the wrong credentials are provided.

Commonly users mistype their password, especially if the passwords are "secure", sometimes fat-fingers happen. (I'll admit I enter my password wrong quite commonly into FreeSO for some reason).

Further more, the way the dialog system was presenting the error messages was on top of the UILoginScreen, so every time a user pressed enter after a wrong password, it would continue to spam the login servers with the wrong credentials.

If the users password is not wrong, maybe the enter their username incorrectly, most users are accustom to being able to press Shift + Tab to return to the previous textfield, FreeSO supported tab into the next field, so adding ShiftTab was relatively straight forward for the Login Dialog.

Thank you for reviewing this PR.

### DIFF
--- a/TSOClient/FSO.UI/Controls/UITextEdit.cs
+++ b/TSOClient/FSO.UI/Controls/UITextEdit.cs
@@ -67,6 +67,7 @@ namespace FSO.Client.UI.Controls
         public event ChangeDelegate OnChange;
         public event KeyPressDelegate OnEnterPress;
         public event KeyPressDelegate OnTabPress;
+        public event KeyPressDelegate OnShiftTabPress;
 
         private UITextEditMode m_Mode = UITextEditMode.Editor;
         private bool m_IsReadOnly = false;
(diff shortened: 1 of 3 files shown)


thertzelle:
I'm not sure why my previous two commits are showing up in this PR, if this could cause problems please reject the PR with that for the reason and I'll find a way to fix my fork.

Thank you.

gmoben:
@thertzelle Looks like your previous PR was squash merged. This one will probably be as well, so it won't matter unless you'd like a clean commit history. In the future, I'd recommend working on a separate branch for each PR/feature you plan to submit.

PR #75, opened 2017-10-04
Try to load iff houses into sandbox mode

francot514:
Added option to load iff houses if TS1Hybrid  path exists.
simshouses


riperiperi:
House**.iff probably isn't the right regex for this, but it works anyways I guess.

francot514:
This code, need to be checked or updated, is giving issues to load houses in latest builds

PR #77, opened 2017-10-30
Roulette EOD

TheArchitectFreeSO:
These are all the changes required for roulette to work except the .piff file for the object.

Users are likely to complain about new players pushing off existing players when attempt to join. The ability for the owner to be the croupier has been removed, as it is useless.

SlotsEOD got a major facelift, as well.


TheArchitectFreeSO:
Here're the piffs, but it's really hard for me to understand why the object seems to reset itself.

The only thing I didn't include in the piff is an updated subroutine to check the machine's balance before allowing it to open (...outside acceptable limits?) since I can't figure out what the object itself resets its attributes mid game (only while open with an NPC croupier).

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODRoulettePlugin.cs
```diff
@@ -0,0 +1,1461 @@
...
+            if (TableBalance < MaxBet * 140)
+                CloseTable();
+            //else
+            {
+                foreach (var player in Players)
+                {
+                    if (player.Client != null)
+                    {
+                        if (player.SimoleonBalance >= MinBet)
+                            player.NewGame(new byte[] { (byte)(MinBet / 255), (byte)(MinBet % 255), (byte)(MaxBet / 255), (byte)(MaxBet % 255),
```
riperiperi: `(byte)(MinBet >> 8), (byte)(MinBet)`
Use bitshifts for short -> byte array, and a cast to byte does %255 (or &0xFF)

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODRoulettePlugin.cs
```diff
@@ -0,0 +1,1461 @@
...
+                else
+                    return false; // invalid amount (should never happen)
+                BetChipCount++;
+                return true;
+            }
+            // can't put more chips on the stack
+            return false;
+        }
+
+        internal bool Decrement(int amount)
```
riperiperi: Could use a dictionary for this. (amount -> bet chip index)

PR #79, opened 2017-11-01
 Roulette Penultimate Patch

TheArchitectFreeSO:
• Machine/Table Balance Textfield and UI Alert Deposit/Withdraw Text
Entry no longer truncate after 5 characters, displaying more than $99999
• Owner Alert now properly displays the minimum balance needed for a
table to open when prompting the owner for their deposit/withdraw amount
• Tables now properly check that the minimum bet is met before accepting
a player's bets. An error is displayed if they player falls short of the
minimum bet
• Roulette plays sounds when a bet is accepted and when a player wins
any money
• "Manage" interaction by owner is now ALWAYS available, which will
prevent tables from becoming stuck and requiring the lot to reset.
Examples of such interruptions include fires and desyncs.
• Multiple tables can be opened and closed and the NPC properly vanishes
upon a table's closing
• A bug was fixed that allowed the owner to set a maximum bet despite
not having enough money in the table to cover it.
• Removed a redundant CloseTable() method call.
• Players cannot play on a table currently being managed by the owner.
• Roommates cannot open a table currently being managed by the owner.

Hopefully the only other patch needed will be once I figure out the subroutine that checks the table's balance.


riperiperi:
LGTM. This fixes the issue where people could not open 2 tables at once, right? (or maybe percieved issue -  it's possible they were running into one of these on the second table)

PR #81, opened 2017-11-02
Roulette Enqueue States

TheArchitectFreeSO:
•No more enumeration/collection changing errors.
•Fixed OnDisconnection
•Thread safe state changes


riperiperi:
Good, should behave a lot nicer now.

PR #84, opened 2017-11-03
Roulette Race Condition

TheArchitectFreeSO:
Moved things around in Croupier:Begin to avoid race condition.


riperiperi:
Since all the stuff in here has kind of been smushed together, I think you should make a fully encompassing PR once Blackjack is completed, then I can review it as a whole. Love the new object tho. 😍

PR #85, opened 2017-11-21
Roulette, Slots CC, & Other minor EOD changes

TheArchitectFreeSO:
• RouletteEOD - Connect to EOD sooner in the Manage (Croupier: Begin) and Play (Player:Begin) trees in order to avoid what appears to be a race condition, where players stand idle for a very long time before connecting to the EOD, sometimes never connecting.
• RouletteEOD - For some reason, MOMI NPCs after the first one do not connect to the EOD. So a failsafe method of checking the object for a connected NPC was implemented. This allows any open tables to be playable, rather than before when the tables assumed they were not open due to a lack of NPC Avatar.
• Minor Volcanic change: This will keep certain TS1 objects from throwing null when trying to check for semi-globals. cf. Blackjack table from TS1
• EOD Timer now works for EOD.TallTall - The textfield and texture/image now appear in the correct spot, and the texture dynamically changes in order to match the expanded/collapsed window.
• EOD.CloseInteraction() - These EODs were not using CloseInteraction()
• RackEOD - If the "Browse" interaction disappears due to an object prim state sync issue (from a sudden lot closure), the owner can make it reappear now by simply choosing "Manage" and exiting the EOD.
• Slot Machine 4 by S1ndle and Raeven


riperiperi:
LGTM. Love the new slot machine 😍

PR #87, opened 2018-01-12
[Server] Allow users to be banned from admin API

ItsSim:
- Make the API able to ban users and disconnect them.
- Show the INV-150 banned message instead of INV-110 message when a banned user tries to login.


ItsSim:
It was indeed a bit random in this pull request, I included it by accident. I have removed all Nancy references and code from my solution though, because I thought all the Api code was being moved to Asp.Net and Nancy would be removed in the future.

I do plan on adding more Api features and doing some control panel stuff soon. Do I need to code for both the Asp.Net and Nancy Apis then? (If Nancy is still going be considered an option?)

riperiperi:
I haven't had time to formally retire it, but it should be retired, yes. The code is still there as a reference implementation, but new features shouldn't be backported for it.

Issue #91, opened 2018-02-02
Login screen exception


Bn8:
I will try to fix it I think I found the root of the problem! I know C# but this is a huge project i'll suggest a pull request if I succeed (its a minor problem anyway), can be reproduced by a+ctrl+delete on textbox

Bn8:
This exception actually happen in EVERY TEXTBOX in the game :(

PR #92, opened 2018-02-02
fix issue #91 exception on login screen using ctrl+delete

Bn8:
Added a m_SBuilder.Length > newEndIndex check so there won't be exception when trying to access index of string beyond bounds.

If I have to use my instincts i'd say the root of the problem is:
SelectionStart = Math.Min(m_SBuilder.Length, SelectionStart);
Where SelectionStart can get the value of the string length, I dont know if its intended (maybe add (-1)?), but i'm not sure so im just making a small hotfix where the problem occurred.


riperiperi:
While bounds checks are always nice, this case should never exist in the first place. It's likely a bug with how the input handler behaves with the control modifier key, so I'd rather fix that on its own.

PR #103, opened 2018-03-22
BlackjackEOD: Table Balance bug fix

TheArchitectFreeSO:
•Fix for a new bug that appeared since the Multi-Tile Object fix for TSOState Balance.
•Clean disconnection added in case a joinable server persists.
•Removed deprecated UI code
•Proper .str.piff allows the viewing of all the stack object's attributes


riperiperi:
LGTM, I'm probably going to be looking at EOD server again when I try to add server self-resync (aimed at resolving and debugging client desync loops).

PR #109, opened 2018-06-29
Developer Experience Improvements

dotequals:
This is just some minor changes to make `git status` less noisy and to prevent the accidental commit of any secrets in `config.json`.

1. In VS2017 I noticed a `.vs` folder being created at the root of the project that I believe can be safely ignored in its entirety.
2. I've removed `config.json` from tracking in favor of `config.sample.json`. The recommended workflow would then be to copy and rename `config.sample.json` to `config.json` and enter your server specific settings there.


dotequals:
I realized that `config.json` wasn't going to get copied over by magic ✨ and adding it to `FSO.Server.csproj` clashes with it being in `.gitignore`.

The solution I came up with was to copy it via the post-build command line and I opted to do this always instead of just on a successful build.

riperiperi:
It just looks like this changes things to be exactly the same. The end result is that a config.json in the project and git tracked (now called config.sample.json) gets copied to the output directory for use in the game.

The main difference would be the action taken when config.json already exists. When I `copy` to an existing file, it asks for keyboard input for overwriting. I don't know what a post build action would do.

The ideal behaviour would be for it to _not_ overwrite and leave the user's customized config.json in the bin folder. Is that what this build action is defaulting to?

dotequals:
`copy` used in a normal command prompt window does ask if you would like to overwrite an existing file, but it seems like that is either being suppressed in VS or `bin` is getting cleared out on every run of Start.

However, if someone places their customized `config.json` at the root of `FSO.Server` now, it will get copied over to `bin` and it won't show up as a modified file that could be accidentally committed.

riperiperi:
Okay, that makes enough sense. For some reason I thought the user might want to put their own config.json in the release folder rather than project directory. Something to watch here might be people trying to add the copied config.json to the project - as since the file would be ignored, it would just cause build errors on copy.

PR #110, opened 2018-07-03
.csproj simplification

dotequals:
Per [this documentation](https://msdn.microsoft.com/en-us/library/ms171454.aspx#Anchor_1), we can use wildcards instead of explicitly defining build actions for files. The `**\*.type` pattern used includes the root folder as well as any recursive sub-directories.

Switching to this method for `.fsom` (Meshes), `.iff` (Objects), and `.piff` (Patches) will make it easier for those creators to add to the game as the steps would be to just drag their file into the proper folder.

I can extend this PR further for other folders and file types that make sense.

EDIT: It looks like I was wrong. If you add the file to the project via Visual Studio it does make an entry in .csproj, but it will still copy the file due to the wildcard entry. We can recommend creators add their files via File Explorer or they can revert the changes to the file via git.


dotequals:
In this PR I've also removed that .vs file that was cached and used the wildcard pattern to copy the JSON files in FSO.Server instead of the post-build step.

dotequals:
Maybe we want to switch the patterns to `Content\MeshReplace\**\*.*`, `Content\Objects\**\*.*`, and `Content\Patch\**\*.*` as I see some images in the MeshReplace folder.

dotequals:
I switched the build action to include every type of file and added the Textures folder as it's possible files may be added to it when creating an object.

PR #112, opened 2018-07-14
Admin API ingame mail fix

ItsSim:
I'm taking a second look at all API-related things I did in January. 
Game mail sending from admin API was wrong, using the user ID instead of the avatar ID.


riperiperi:
Why is there a change to FSOMonogame in here? I'm guessing it changes the submodule commit?

TheArchitectFreeSO:
I experienced the same thing with Github but chose to ignore this change:
https://i.gyazo.com/99bf9624ab40505276cb9511592b902a.png

ItsSim:
Oops! Guess that should work.

riperiperi:
LGTM, weird that I didn’t see this during the review. Note this probably wont be deployed on production til next update, since current master has a small problem with sim to sim interactions.

PR #113, opened 2018-07-18
Final Proof of Concept: Custom Content — HoldemCasinoEOD

TheArchitectFreeSO:
**HoldemCasinoEOD**
It's a modern casino version of the classic Texas style tabletop card game. Rather than having the best 5-card hand at the table to win the pot, each player must simply have a better 5-card hand than the dealer in order to win big!

Players must place an ANTE (or blind) BET to be dealt in. They may also place an optional SIDE BET. Once the betting round ends each player is dealt 2 hole cards face up, the dealer is dealt 2 hole cards face down, and then 3 community cards are laid on the table face up (the flop). Community cards are shared between all players and the dealer but, unlike Texas Hold'em, players only need to beat the dealer as their neighbors' hand doesn't matter.

As soon as the flop happens, any player who made a side bet can receive a payout based on their 5 card hand (2 hole + 3 community) but ONLY if it contains a Pair of Aces or better (2 Pair, Trips, Flush, etc.). After side bet payouts, with the dealer's hole cards still face down, each player gets a turn (starting from dealer's left) to decide if they will CALL or FOLD.

If a player folds they surrender their ante bet. But if they wish to call, they must put down an additional CALL BET, which is always 2 times their ante bet. After at least 1 player calls, 2 more community cards are laid down face up on the table (the turn and the river). 

Finally the dealer turns over their 2 hole cards and each hand is scored based on the best possible 5 card hand of the 7 cards available (2 hole + 5 community).

If the PLAYER has the better hand:
The player receives a payout of their ante bet based on the type of 5 card hand they have.
In Addition:
•If the dealer qualifies—has a pair of 4's or better—the player wins a 1:1 payout of their call bet
•If the dealer doesn't qualify—has a pair of 2's, 3's or just a High Card—the call bet is pushed, i.e. returned to the player.

If the DEALER has the better hand:
The player loses both the ante and call bets.

If the PLAYER and DEALER have equal hands:
The Ante and Call bets are pushed, i.e. returned to player

NOTE: Remember, the winner will have the BEST 5 CARD HAND of the available 7 cards. That means if the player and the dealer both have 2 Pair (2's and 4's), the winner MIGHT be determined by the second hole card that the player or dealer has.

**Additional Changes**
[FSO.IDE]
•Strings: putting $//n into a string will be parsed to a single new line

[RouletteEOD]
•A special check was added before every new game to make sure the object didn't break. If the object broke, all players and the dealer NPC will be booted and a message will tell the players that the table broke.

[BlackjackEOD]
•The dealing cards effect in the UIEOD was hastened to match the faster animation speeds.
•A special check was added before every new game to make sure the object didn't break. If the object broke, all players and the dealer NPC will be booted and a message will tell the players that the table broke.


TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODHoldEmCasinoPlugin.cs
```diff
@@ -0,0 +1,2187 @@
...
+            }
+            bool isOwner = (((VMTSOObjectState)Server.Object.TSOState).OwnerID == client.Avatar.PersistID);
+            // if the client is the owner of the object AND there were no failures detected
+            if ((isOwner) && (failureReason.Length == 0))
+            {
+                // attempt to credit the owner by debiting the object
+                var VM = client.vm;
+                VM.GlobalLink.PerformTransaction(VM, false, Server.Object.PersistID, client.Avatar.PersistID, withdrawAmount,
+
+                (bool success, int transferAmount, uint uid1, uint budget1, uint uid2, uint budget2) =>
```
riperiperi: Weird newline?

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODHoldEmCasinoPlugin.cs
```diff
@@ -0,0 +1,2187 @@
...
+                if (failureReason.Length == 0)
+                    failureReason = VMEODSlotsInputErrorTypes.InvalidOwner.ToString();
+                client.Send("holdemcasino_deposit_fail", failureReason);
+            }
+        }
+        #endregion // dealer events
+        #region Events
+        // client closes UI
+        private void UIClosedHandler(string evt, byte[] data, VMEODClient client)
+        {
```
riperiperi: What does data contain here?

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODHoldEmCasinoPlugin.cs
```diff
@@ -0,0 +1,2187 @@
...
+                                // enqueue the sequence to debit the player
+                                slot.PendingTransaction = true;
+                                EnqueueSequence(client, VMEODHoldEmCasinoEvents.Player_Place_Bet, newAnteBet + newSideBet);
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        // this semantics event occurs after any animation sequence during gamestate = entre'act
```
riperiperi: *sim*antics!!!

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODHoldEmCasinoPlugin.cs
```diff
@@ -0,0 +1,2187 @@
...
+            }
+        }
+        /*
+         * [WORST CASE SCENARIO ANTE PAYOUT]
+         * Ante Bet (Pay Table 2) => Community cards would be { Ten, Jack, Queen } suited and { 7 or below any suit, same number any suit }
+         * Example community cards { 10c, Jc, Qc, 7s, 7h }, Dealer Qualifies (pair of fours or better) but loses against all players
+         ***********************************************************************************************************************************
+         * Player# - [Type of Hand] - Payout ratio - (desc. of hand) e.g. example
+         ***********************************************************************************************************************************
+         * Player1 - [Royal Flush] - 25:1 - (has King and Ace of same suit as three communit
```
riperiperi: 👌 love the detailed comments

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODHoldEmCasinoPlugin.cs
```diff
@@ -0,0 +1,2187 @@
...
+                                            if (Dealer != null)
+                                            {
+                                                if (GameState.Equals(VMEODHoldEmCasinoStates.Waiting_For_Player) || GameState.Equals(VMEODHoldEmCasinoStates.Closed))
+                                                    EnqueueGotoState(VMEODHoldEmCasinoStates.Betting_Round);
+                                                else // player is joining mid-game or mid-betting round
+                                                {
+                                                    if (GameState.Equals(VMEODHoldEmCasinoStates.Betting_Round))
+                                                    {
+                                                        // sync all ante and side bets
+                                                        List<string> acceptedBets = GetAllAcceptedBets();
```
riperiperi: This is not thread safe as it iterates Lobby.Players, which can change on the main thread.
As a rule, anything that isn't a single primitive number could partially change when running on another thread. Primitive number types change atomically, so your problem there is thinking if multiple could change that depend on each other's value remaining consistent.

This one is quite tricky to solve. I'm OK with you ignoring this, but it would be nicer if you somehow made these actions happen on the main thread. (usually I use as lock guarded List<Action> to run arbitrary functions on the main thread)

riperiperi: In fact, this entire PerformTransaction handler could use some running on the main thread, it's very hard to reason about this much access to shared state. Might be worth considering a standard way to do this between all EOD servers using helper functions.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODHoldEmCasinoPlugin.cs
riperiperi: delet this

riperiperi: Wouldn't this work? (similar for the examples below, though enumeration means you have to use these functions on main thread)
`Players.All(x => (x == null) || Lobby.GetSlotData(x).BetSubmitted)`

riperiperi: Same
`Players.All(x => (x == null) || Lobby.GetSlotData(x).PendingTransaction)`

riperiperi: Same
`Players.Any(x => (x != null) && Lobby.GetSlotData(x).BetAccepted)`

TSOClient/tso.client/UI/Panels/EODs/UIHoldEmCasinoEOD.cs
```diff
@@ -0,0 +1,2125 @@
...
+                        }
+                    }
+                    else
+                    {
+                        // if the first char of the first string is 'q' then the dealer didn't qualify
+                        if (winningsString[0] == 'q')
+                        {
+                            winningsString = winningsString.Remove(0, 1);
+                            // "Dealer did not quality. You win $% d"
+                            SetNewTip(GameFacade.Strings.GetString("f111", "72").Replace("%d", winningsString) + ".");
```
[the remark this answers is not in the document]
TheArchitectFreeSO: "GameFacade.Strings.GetString("f111", "72", new string[] {winningsString});"
Very cool, did not know this one.

TSOClient/tso.client/UI/Panels/EODs/UIHoldEmCasinoEOD.cs
```diff
@@ -0,0 +1,2125 @@
...
+        {
+            if (newBetArray != null)
+            {
+                var newBet = VMEODGameCompDrawACardData.DeserializeStrings(newBetArray);
+                if (newBet != null && newBet.Length > 0)
+                {
+                    var anteBet = newBet[0];
+                    if (anteBet.Equals("0") || anteBet.Equals(""))
+                        anteBet = "Ante";
+                    var playerIndex = (int)Char.GetNumericValue(evt, evt.Length - 1);
```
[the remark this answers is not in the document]
TheArchitectFreeSO: This works, but what am I subtracting? Is the difference here between the int value of the letter minus the int value of char zero?

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODHoldEmCasinoPlugin.cs
TheArchitectFreeSO: Definitely agree, fixes all around.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODHoldEmCasinoPlugin.cs
TheArchitectFreeSO: Comment added, basically player vs owner

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODHoldEmCasinoPlugin.cs
TheArchitectFreeSO: Fixed, ironically due to semantics

TSOClient/tso.client/UI/Panels/EODs/UIHoldEmCasinoEOD.cs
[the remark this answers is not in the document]
riperiperi: Nope, it calculates the `char` difference and converts that into an int. That would be the character's offset from `0`, which is 0 for `0`, 1 for `1` etc. Note you'd have to bounds check this if you didn't trust the output.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODHoldEmCasinoPlugin.cs
riperiperi: What would happen if a player said they were an owner?

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODHoldEmCasinoPlugin.cs
```diff
@@ -64,13 +65,22 @@ public VMEODHoldEmCasinoPlugin(VMEODServer server) : base(server)
...
+            SyncQueue = new List<VMEODClient>();
         }
         #region Public
         public override void Tick()
         {
             if (Controller == null)
                 return;
 
+            // sync any latecomers' UIEODs
+            if (SyncQueue.Count > 0)
```
riperiperi: Wrap this in `lock (SyncQueue) {`, and use `SyncQueue.Clear()` (need the reference to not change for locking reasons)

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODHoldEmCasinoPlugin.cs
```diff
@@ -286,15 +306,14 @@ public override void OnConnection(VMEODClient client)
...
-                                                        if (acceptedBets != null)
+                                                        // send all ante and side bets
+                                                        if (acceptedBets != null && acceptedBets.Count > 0)
                                                             client.Send("holdemcasino_sync_accepted_bets", VMEODGameCompDrawACardData.SerializeStrings(acceptedBets.ToArray()));
                                                         client.Send("holdemcasino_toggle_betting", new byte[] { 1 }); // "Place your bets..."
                                                     }
                                                     else
                                                     {
-                                                        SyncAllPlayers(client);
+                                                        SyncQueue.Add(client);
```
riperiperi: `lock (SyncQueue) SyncQueue.Add(client);`

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODHoldEmCasinoPlugin.cs
TheArchitectFreeSO: If they did that but were actually a player, then the dealer would be removed and the table would close anyway.

PR #114, opened 2018-07-20
Fixes 3D shortcut limitations

RyanSmake:
There is no reason to my understanding why these should not work in 3D.


riperiperi:
LGTM. This was likely a mistake.

PR #115, opened 2018-07-21
 Save last_login when users login through AuthLoginController

ItsSim:
I don't think this data is obtainable from anywhere in the database at this time. I think it can be pretty useful.


riperiperi:
LGTM. Note that the next update will move our ASP.NET API to ASP.NET core, so I'll be porting this and future updates to the API too.

PR #116, opened 2018-08-01
TwoPersonJobObjectMazeEOD

TheArchitectFreeSO:
With a twist...
Also UIImages can now have tooltips


TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODTwoPersonJobObjectMazePlugin.cs
```diff
@@ -0,0 +1,502 @@
...
+            byte color = (byte)CharismaPlayerOrigin.CellData.Color;
+            CharismaPlayer.Send(eventName, new byte[] { wallConfig, color });
+        }
+        private void SendLogicPlayerData()
+        {
+            if (LogicPlayer != null)
+            {
+                LogicPlayerHasInit = true;
+                
+                // get the wall configs for half of the cells: every other one in each row, but staggered starting at 0 or 1 moving down rows
```
riperiperi: This is a really weird way of doing this - I would have sent only the top&left wall of each tile. I guess this compresses better.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODTwoPersonJobObjectMazePlugin.cs
```diff
@@ -0,0 +1,502 @@
...
+            colorIndex = Rng.Next(0, cellList.Count);
+            CharismaPlayerOrigin = cellList[colorIndex];
+
+            // get the solution
+            var numberOfMoves = Rng.Next(MIN_MOVES_TO_EXIT, MAX_MOVES_TO_EXIT + 1);
+            SolutionPath = CurrentMaze.GetPathFromOriginToExit(CharismaPlayerOrigin, null, numberOfMoves);
+            if (SolutionPath != null)
+                SolutionCell = SolutionPath[SolutionPath.Count - 1];
+            else
+                Reset();
```
riperiperi: Possible `StackOverflowException` if we get really bad luck. It is incredibly unlikely but they can bring down the whole server, as stack overflows cannot be caught. I'll allow it for now given the circumstances, it's just a future concern. Example:

`Reset()->CurrentMaze.BuildFromOrigin()->(OnMazeGenerated)MazeGeneratedHandler()->Reset()->CurrentMaze.BuildFromOrigin()->...`

It may be better to have a loop in Reset() that continues while a maze has not been found. With impossibly bad RNG this will still stall but won't crash the server (and we have a chance to exit if things are getting bad). In fact, you can check this right after CurrentMaze.BuildFromOrigin().

PR #120, opened 2018-10-19
DiscordRpcEngine Updates, Chat Features, Fixes

AstroPsynapse:
**DiscordRpcEngine-related:**

[FSO.UI.Model.DiscordRpcEngine]
- Added an additional method for updating Discord rich presence to allow other game screens to be shown (Main Menu, CAS/SAS, etc)
- Make rich presence respect ingame privacy (Where Am I?) settings
- Added information about which sim is currently playing.
- Tidied up strings a little bit.
- Updated job information to include the Nightclub.

[FSO.Client.Controllers.GizmoController]
- Added a call to DiscordRpcEngine.SendFSOPresence() when avatar information is received by the gizmo.**
_** I'm not sure if this is the best place for this, but I wasn't sure on where else I can put it. It seemed sensible to do it here since it calls for the avatar data upon loading and doesn't change._

[FSO.Client.GameController]
- Added various calls to DiscordRpcEngine.SendFSOPresence() to show what screen the user is in.

[FSO.Client.UI.Panels.UIPersonPage]
- Update Discord rich presence when privacy mode is changed.

[FSO.Client.UI.Screens.CoreGameScreen]
- Updated Discord rich presence calls to respect privacy mode and include sim name.

[FSO.Client.UI.Screens.PersonSelectionEdit]
- Added call to DiscordRpcEngine.SendFSOPresence() to display "In Create A Sim"**
_** This was placed here in order to detect CAS in both online and sandbox mode._

**Chat:**

[FSO.Client.GlobalSettings]
- Added configuration entries for chat timestamp, size, text size, location, and opacity.

[FSO.Client.Controllers.MessagingController]
[FSO.Client.Controllers.MessagingWindowController]
- Add ChangeOpacity method

[FSO.Client.UI.Panels.UIChatDialog]
- Get chat text size, window location, window size, opacity from configuration entry.
- Saves configuration entries after window is resized/moved/etc.
- Added chat timestamps support (configurable from the chat panel)
- Increased upper bounds on text resizing so it can go slightly bigger than before.

[FSO.Client.UI.Panels.UIMessageWindow]
- Changed container type to UICachedContainer to allow opacity changes for private message windows**
_** UIContainer (non-cached) wasn't properly supporting opacity changes which is why this was changed._
- Add update hook to listen for opacity changes.

[FSO.Client.UI.UIOptions]
- Added chat opacity and timestamp options.
- Renamed TTS options horizontal container variable.

[FSO.SimAntics.NetPlay.Model.VMChatEvent]
- Added Timestamp property for chat timestamp support.

[Strings]
- Added new strings in _f113_chatexpansionstrings.cst for chat opacity and timestamp panel controls.

**Fixes:**

[FSO.Client.UI.Controls.UIVMPersonButton]
- Fixed jumping to a sim when right click scrolling over a sim's panel button.

**Other:**

[FSO.Client.Utils.GameLocator.WindowsLocator]
- Added check for local directory after registry check. If not found, fall back to default install location


TSOClient/tso.client/UI/Controls/UIVMPersonButton.cs
```diff
@@ -66,7 +67,8 @@ public override void Update(UpdateState state)
                 UpdateAvatarState(perm);
             }
             RMB = state.MouseState.RightButton == ButtonState.Pressed;
-            if (RMB && Hovered) CenterPerson(this);
+            RMBscroll = GameFacade.Cursor.CurrentCursor != Common.Rendering.Framework.CursorType.Normal;
```
riperiperi: That's kind of a hack, it's probably just better setting a local variable `rmb` to the current click state then checking `RMB && !rmb` to see if the right mouse button was clicked (wasn't down last frame, is down now) and then checking if the mouse position is on our bounds. Set private `RMB = rmb` afterwards so you store the last state we saw.

riperiperi:
Will merge for now since I'm pushing an update.

PR #122, opened 2018-10-26
Tiny EOD patch

TheArchitectFreeSO:
•A small null check added due to a yet unresolved (rare) bug with BandEOD where a Sim's client disconnects from the EODServer but the EODPanel doesn't close. Manually closing the panel after this caused a null error, which is what this patches (UIEOD.cs).
•A seasonal costume was added for both genders to the "costume" trunk
•A new type of EODPanel has been added for future EODs


riperiperi:
The things we do for events. 🙄 

I'll let it through for now, but as discussed on discord I'd rather a way to do this that's less hardcoded, and also does not involve modding TSO files.

PR #124, opened 2018-10-27
Fix decimal numbers for different User Culture settings

AstroPsynapse:
Fixes the issue with default config values that contain decimals (floats, decimals, doubles). It should also update accordingly if a user should ever switch their number separator format.


AstroPsynapse:
That makes a lot more sense, otherwise the configuration could get messy depending on different formats. 

I changed it to save and load values using `CultureInfo.InvariantCulture` only.

PR #125, opened 2018-11-03
FOV changer on tab camera

jwofles:
Hi this was really confusing and im sorry i screwed up the lines.
I probably done this wrong but I'm just submitting it anyway 👍
https://cdn.discordapp.com/attachments/267027937197424641/508014333624582155/2018-11-02_20-26-29.gif
edit - i realised i changed this "rcState.RotationX -= (mpos.X - mx) / 500f;" to 100f by accident my bad


riperiperi:
Why did you change the FOV to be a divisor of Math.PI? Before the FOV was an angle in radians, which is the most reasonable format.

jwofles:
> Why did you change the FOV to be a divisor of Math.PI? Before the FOV was an angle in radians, which is the most reasonable format.

I don't know. I didn't know what radians were back then (wish I was kidding).

PR #126, opened 2018-11-30
Moderate Fridge Fix

dotequals:
This patch allows `Fridge - Moderate` to be repaired by visitors like the others


riperiperi:
Interesting commit history, but I suppose Github will squash it anyways. LGTM.

PR #127, opened 2018-12-13
Right-Click Copy GUID to Clipboard

dotequals:
This change allows you to copy an object's GUID to the clipboard by right-clicking the GUID button on the Object tab of the Edit Object window. It currently does this silently, but could update the GUID label to something like GUID (Copied!) to make it more user friendly.

I also opted for right-click in case there's already a plan for left-click.


dotequals:
Including this in a more comprehensive PR that uses WinFormsClipboard instead of reinventing the wheel.

PR #128, opened 2018-12-13
XML Entry Tab and Right-Click Copy GUID

dotequals:
image

This PR adds an XML Entry tab to generate catalog_downloads.xml entries which can then be copied and pasted into the game's XML file. It also includes some double checking such as trying to put a free item into any category other than debug.

It also includes the feature from #127, but now uses WinFormsClipboard.


riperiperi:
LGTM, though this is rather situational for it to be a main tab in the object window. Perhaps I can find some other place for it later.

PR #129, opened 2018-12-19
Makes Sim's head look at Pie Menu Buttons

LazyDuchess:
My C# is not the best but, thought i'd make a small contribution anyways
head2


riperiperi:
Love it, actually looks better than TS1's too.

PR #130, opened 2018-12-31
Hex Editing for Primitives with Unknown Operands

dotequals:
image

This adds the ability to just edit the hex values of any primitives that don't have overrides.

Currently it doesn't pad zeros, but maybe it should.


dotequals:
Validation doesn't seem to always work properly. Creating a new Inventory Operation and setting Argument 1 to `abcdef` crashes and sets the value to 8000 instead of back to 0.

riperiperi:
I guess I can look into validation (surely winforms should handle this), but this is definitely better than nothing. LGTM.

PR #131, opened 2019-01-28
Adjacent resources

dotequals:
This is kind of an opinionated change, but I'd like to separate the TSO folder from being inside the FreeSO folder portable-type installs. In the event FreeSO troubles arise, people may blow away their TSO data without realizing.

I'd also be okay with removing all the other non-standard places to look, but didn't want to cause a breaking change 👼

Also 🤫 the commit history, it keeps happening


TSOClient/tso.client/Utils/GameLocator/WindowsLocator.cs
```diff
@@ -15,6 +15,10 @@ public string FindTheSimsOnline()
         {
             string Software = "";
 
+            // Search relative directory similar to how macOS and Linux works; allows portability
+            string localDir = @"../The Sims Online\TSOClient\";
```
riperiperi: `/` and then `\`? Then you replace it later anyways. For best cross platform support, we always use `/` since it also works on windows.

PR #134, opened 2019-03-15
Fix AV in Simitone.

simptomo:


riperiperi:
I'm not sure what the PR title means, but this issue was due to Tuning not existing in TS1 mode. I solved it a few weeks ago by adding a null propagation operator to tuning, since it already dealt with null values by setting them to 0. Granted this only exists so I can toggle off bezier routing in TSO for debug purposes.

https://github.com/riperiperi/FreeSO/commit/bbae051a782948ab2069119f41170e85579784f6

PR #135, opened 2019-03-16
Fix: People cannot call phone actions in Simitone.

simptomo:
Thanks!

### DIFF
--- a/TSOClient/tso.content/TS1/TS1ObjectProvider.cs
+++ b/TSOClient/tso.content/TS1/TS1ObjectProvider.cs
@@ -54,7 +54,7 @@ namespace FSO.Content.TS1
                             Source = GameObjectSource.Far,
                             Group = (short)obj.MasterID,
                             SubIndex = obj.SubIndex,
-                            GlobalSimObject = obj.ObjectType == OBJDType.SimType && obj.Global == 1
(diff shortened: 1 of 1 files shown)


riperiperi:
Hey,

Thanks for the PRs, If I knew someone was going to work on Simitone I would have committed my local changes to fix things up! I've done that now, but you might have to manually set FreeSO's submodule commit to the latest.

As for this change, you appear to have made every object in the game a GlobalSimObject. This will spawn every "normal" object in the game out of world whenever you load a property - that is very slow and could cause some very unexpected things to happen. I haven't encountered this issue with the latest changes I committed, so perhaps see if the issue persists on those?

I'll be working on FreeSO for a bit longer so these changes I committed will be the last for a while - meaning any fixes you want to work on will go unimpeded, but I recommend the following for future PRs:

- Need more detailed commit/PR messages. You should be explaining what the issue was in the first place, and how your change fixed it. This is to make things as easy as possible to review.
- You should discuss changes on the discord before making a PR for them. If it's simple enough then you're fine not doing that, but if they have far reaching implications like this PR then it's best to figure out how these things should work first.

riperiperi:
Addressed by myself - There is one global object that is *not* normal type and should be included, and many of type 12 (gift tokens?), so I made the check `GlobalSimObject = obj.Global == 1 && obj.ObjectType != OBJDType.GiftToken`

PR #138, opened 2019-03-28
[Fix for Simitone]My dogs cannot live.

simptomo:


riperiperi:
This change would make TS1AllowDogs the same flag as AutoFirstSelect, which is in TSO and TS1. How did you verify that this was flag 1<<8?

simptomo:
That's what I want to know.

riperiperi:
Identified issue: TS1AllowDog and TS1AllowCat flags were overlapping with some TSO ones, which was causing weird engine behaviour. Will be fixed in upcoming version.

PR #141, opened 2019-03-29
[Fix for Simitone] Half neighbour generated.

simptomo:


riperiperi:
I'm assuming this happens if there are not enough entries in the STR? Why not ensure that and add the entries manually, rather than trying to set the string then checking if it didn't set?

An easy way to do this would be adding empty strings to the STR to pad it to the required length before we replace the bodystrings. A smarter way would be to replace the ones in the existing range, and add the ones outwith it.

riperiperi:
I'm closing this since there wasn't really any feedback.

PR #143, opened 2019-04-12
[Fix for Simitone]Some custom objects get an error.

simptomo:


riperiperi:
Not a fan of the double ternary, and the default value should probably be `entry.AttenuationValue`. Can you change it to:
```csharp
float atten = entry.AttenuationValue;
if (entry.AttenuationCode > 0 && entry.AttenuationCode < TTAB.AttenuationValues.Length) 
{
    atten = TTAB.AttenuationValues[entry.AttenuationCode];
}
```

riperiperi:
FindBestAction has been reworked, but I included a fix like this,

PR #144, opened 2019-04-12
[Fix4simitone] I want to sit on The Polar Bear Rug!

simptomo:
The Polar Bear Rug ↓
キャプチャ


riperiperi:
Nice catch, looks like this only happens when the object it's searching for isn't on the list.

PR #146, opened 2019-04-15
[Fix for Simitone] Some objects are gone.

simptomo:
Before:
キャプチャBad

After:
キャプチャgood


riperiperi:
image
The version of this lot for Complete Collection seems to have a few more objects (eg. the flowers on the table, dog bowls), even without this change. Can you verify that there are no more missing objects by looking at the property in TS1?

riperiperi:
Will merge for now. Would like further investigation on this tho.

simptomo:
You are right.

See https://youtu.be/OlFayNExyzM?t=336 or https://sims.fandom.com/wiki/Al_Pine%27s_Winter_Wonderland

I will check too.

PR #147, opened 2019-04-16
[Fix for Simitone] Neighborhood.iff will be destroyed.

simptomo:


riperiperi:
Wow, I really screwed that one up. LGTM

PR #148, opened 2019-05-19
Press I to inventory an item, while it's being held.

thertzelle:
Uses previously existing function which checks:
- Is your focus currently in another textbox? (Previous Deletion Commit Addition)
- Are you holding it?
- Do you own it?
- Can you delete it?

Tested in Sandbox, works well !


riperiperi:
LGTM.

PR #153, opened 2019-08-18
[API] Added funtionality and efficiency

Cowplant-Simmer-Collin:
- userOauth token (with specified permissions)
- avatarinfo
  get avatars from specific user (req user auth)
  get avatars from multiple ids
  get all avatars now can get up to 500 avatars per page
  getting all neighborhood avatars is more efficient
  online avatars can also be requested as compact now (only online count)
- lotinfo
  json objects now include lot_id (more consistent/reliable value than location)
  json objects now include admit_mode
  get lots from multiple ids
  online lots can also be requested as compact now (only online count)
  added funtion for getting all the lots by pages, and max 500 lots per page

### DIFF
--- a/TSOClient/FSO.Server.Api.Core/Controllers/UserOAuthController.cs
+++ b/TSOClient/FSO.Server.Api.Core/Controllers/UserOAuthController.cs
@@ -1,0 +1,128 @@
+﻿using FSO.Server.Api.Core.Utils;
+using FSO.Server.Common;
+using FSO.Server.Servers.Api.JsonWebToken;
+using Microsoft.AspNetCore.Cors;
+using Microsoft.AspNetCore.Mvc;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Web;
+using System.Web.Http;
+
+namespace FSO.Server.Api.Core.Controllers
+{
+    [Route("userapi/oauth/token")]
+    [ApiController]
+    public class UserOAuthController : ControllerBase
+    {
+        [HttpPost]
+        public IActionResult CreateToken([FromForm] UserOAuthRequest userAuthRequest)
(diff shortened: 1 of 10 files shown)


TSOClient/FSO.Server.Api.Core/Controllers/UserOAuthController.cs
```diff
@@ -0,0 +1,128 @@
...
+                    }
+                    return ApiResponse.Json(System.Net.HttpStatusCode.OK, new UserOAuthError("unauthorized_client", "user_credentials_invalid"));
+                }
+
+                //user passed the password check, and now creates the claim/token
+                da.Users.SuccessfulAuth(user.user_id, ip);
+                var claims = new List<string>();
+
+                //set the permission level in the claim
+                switch (userAuthRequest.permission_level)
```
riperiperi: What will be the extent of these permissions? They seem to be granted immediately to any user authenticating, and obviously right now the only one that works is reading avatars for the logged user.

TSOClient/FSO.Server.Api.Core/Controllers/UserOAuthController.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Web;
+using System.Web.Http;
+
+namespace FSO.Server.Api.Core.Controllers
+{
+    [Route("userapi/oauth/token")]
+    [ApiController]
+    public class UserOAuthController : ControllerBase
```
riperiperi: VERY IMPORTANT: make sure this auth does not allow admin actions to be performed. (admin api) Since it doesn't add the "admin" or "moderator" claims, it will probably be fine, but it definitely would not hurt to check.

riperiperi:
LGTM, thanks for addressing the feedback!

PR #154, opened 2019-08-31
Support of Cyrillic, Greek and extended latin

Cosmatevs:
- slightly improved existing characters, recreated some
- recreated diacritics and composed characters of Latin-1 Supplement
- created monotonic Greek and Cyrillic characters used in most spoken languages
- redone spacing
- painfully redone kerning
- made numbers monospaced (because of dynamic labels)
- removed Japanese and Chinese characters, and gothic characters, since they're unused
font
I am not a professional font designer but I think the effect is rather pleasing.
Unfortunately, I couldn't see how it looks in the game.
Tell me if you need any glyphs. I'll see what I can do.


riperiperi:
Looks very good! Sorry for the delay, must have missed the email. Will look at the new characters and rebuild the msdf font for ingame.

riperiperi:
I noticed you removed characters from the font including some of the symbols. I'd prefer if they were only added - not removed - as if they are not needed they can simply be excluded from the ini. Symbols like ┛, ♪ were used in the ini, and I notice the only one you didn't remove is the stars (because it it used by upgrades.

The ttf is not distributed with the game (we distribute a font resource generated with the ini config), so the extra size is not an issue. Recently I added support for dynamically loadable "extension" character sets, which the japanese characters in this font could still be used for, rather than pulling an additional font.

The glyphs you've added are brilliant - they look indistinguishable from the maxis ones in terms of style, which is as perfect as you can ask for.

Cosmatevs:
I've kept non-gothic symbols only, and recreated stars. I'll add again all symbols for now, and recreate them later if they're going to be used.

Japanese characters don't follow the font style and they could be difficult to read in small sizes since they're quite thin. For that reason, I think Japanese should have its own font, like Thai has. By the way, I've deleted Japanese characters because it takes a few seconds to save such a big file and it was some inconvenience for me saving a file frequently. I could add Japanese characters again if you want.

And thanks for your kind words, I tried my best.

Cosmatevs:
I've just added Vietnamese glyphs but double-diacritized capital letters are so high that I'm afraid they're going to be cut.
vietnamese

riperiperi:
The msdf font generator rescales the characters to fit in the box, so big characters should just lose fidelity rather than clipping.

On 17 Sep 2019, at 21:23, Cosmatevs <notifications@github.com<mailto:notifications@github.com>> wrote:


I've just added Vietnamese glyphs but double-diacritized capital letters are so high that I'm afraid they're going to be cut.
[vietnamese]<https://user-images.githubusercontent.com/44393502/65076683-c29fff80-d999-11e9-9ee3-51a2a81b973f.png>

Cosmatevs:
Thus some Vietnamese characters will be rescaled…

I'll probably be changing/extending this font in the future but for now that's all. Consider using the font I've mentioned before, as a fallback for Japanese characters.

riperiperi:
It's not really something to consider since the difference is that miniscule. For example, in the english font , has a lot more "resolution per distance" than like, W. I'll give this version a run through the content builder and test it ingame.

TSOClient/tso.content/ContentSrc/Fonts/simdialogue.ini
```diff
@@ -1,6 +1,6 @@
...
\ No newline at end of file
+ranges=( ,~),( ,±),(´,¸),(º,»),(¿,ň),(Ŋ,ſ),(Ə,Ə),(Ơ,ơ),(Ư,ư)(Ș,ț),(Ȳ,ȳ),(ȷ,ȷ),(ə,ə),(ˆ,ˇ),(˘,˝),(̉,̉),(̒,̓),(̛,̛),(̣,̣),(̦,̦),(;,;),(΄,Ί),(Ό,Ό),(Ύ,Ρ),(Σ,ώ),(Ѐ,џ),(Ґ,ғ),(Җ,ҝ),(Ҡ,ң),(Ҫ,ҫ),(Ү,ҳ),(Ҷ,һ),(Ӏ,ӂ),(ӏ,ӑ),(Ӗ,ә),(Ӣ,ӣ),(Ө,ө),(Ӯ,ӯ),(Ӳ,ӳ),(Ạ,ỹ),(‐,―),(‘,‣),(‥,…),(‰,‰),(′,″),(‹,※),(€,€),(⃘,⃘),(℃,℃),(№,№),(℡,™),(Å,Å),(Ⅰ,ⅿ),(←,↕),(⇒,⇒),(⇔,⇔),(∀,∀),(∂,∃),(∇,∈),(∋,∋),(∑,∑),(√,√),(⇒,⇒),(⇔,⇔),(∀,∀),(∂,∃),(∇,∈),(∋,∋),(∑,∑),(√,√),(∝,∠),(∥,∥),(∧,∬),(∮,∮),(∴,∵),(∽,∽),(≈,≈),(≒,≒),(≠,≡),(≤,≧),(≪,≫),(⊂,⊃),(⊆,⊇),(⊥,⊥),(⊿,⊿),(⌒,⌒),(①,⑳),(─,┃),(┌,┌),(┏,┏),(┐,┐),(┓,└),(┗,┘),(┛,┝),(┠,┠),(┣,┥),(┨,┨),(┫,┬),(┯,┰),(┳,┴),(┷,┸),(┻,┼),(┿,┿),(╂,╂),(╋,╋),(■,□),(▲,△),(▼,▽),(◆,◇),(◊,○),(◎,●),(◯,◯),(★,☆),(♀,♀),(♂,♂),(♪,♪),(♭,♭),(♯,♯),(　,〕),(〝,〟),(,)
```
riperiperi: Missing comma here: `(Ư,ư)(Ș,ț)`

The MSDFExtension also has trouble importing this file, but that was on my end and will be fixed with a commit to master. For some reason it was forcing read as ASCII rather than UTF-8 (how did thai characters work?)

riperiperi:
Did some ingame testing, looking very good.

Vietnamese example:
image

Double diacritized:
image

New star on upgrades:
image

Cyrillic:
image

Large Scaling (signed distance field upscaling):
image

TSOClient/tso.content/ContentSrc/Fonts/simdialogue.ini
Cosmatevs: Added the missing comma and removed a private range that I guess causes problems. But I don't know how the Thai characters will work now since they contain references to characters in the private range.

Cosmatevs:
Cyrillic characters aren't kerned well, from what I see. The font has got three kerning tables – for Latin, Cyrillic and Greek. I'm afraid that the first one only works… By the way, I've just found an unkerned pair.

riperiperi:
The truetype library that I am using reads one kern-table that can read multiple subtables. 

https://github.com/roy-t/TrueType/blob/master/RoyT.TrueType/TrueTypeFont.cs
https://github.com/roy-t/TrueType/blob/master/RoyT.TrueType/Helpers/KerningHelper.cs

If multiple subtables are used, the kerning should still be detected.

Cosmatevs:
But probably it isn't detected at all, even the Latin part. Look at `To` in _Touch_, or `ГД`.

riperiperi:
Not sure how that can be the case, as the kerning I added to the font originally did work.

riperiperi:
image
Compiling it is showing 0 kerning pairs, but I cannot debug the MSDFExtension due to how MGCB works.

Cosmatevs:
I've saved the font using the _old style 'kern'_ option. Does it work now?

riperiperi:
image
Yep. Hopefully C# dictionary is fast enough for this.

image

riperiperi:
LGTM, thank you so much! I'll commit the change to the MSDF extension and a compiled version separately.

Issue #155, opened 2019-09-20
Unexpected behaviors while typing using AltGr


Cosmatevs:
Many keyboard layouts use `AltGr` to allow to type diacritized letters or special characters. FreeSO doesn't take it into account for now. `AltGr`+`A` erases whole message you're typing, `AltGr`+`S` opens the lot save dialog, etc.

![add-a-page](https://user-images.githubusercontent.com/44393502/65344244-780bc680-dbd7-11e9-825b-8a9980a4242e.gif)

Cosmatevs:
Reason:
`KeyboardState.IsKeyDown(Keys.LeftControl)` is true when AltGr is down. So, for example, `AltGr`+`A` selects whole message and then `ą` appears and replaces the message.

PR #156, opened 2019-09-21
Makeshift fix for AltGr treated as Ctrl

Cosmatevs:
A makeshift solution to #155 – AFAIK, you cannot catch the difference between `LeftCtrl`+`RightAlt` and `RightAlt`.


riperiperi:
Good catch, looks like this is an issue with Monogame (mentioned in this thread):

https://github.com/MonoGame/MonoGame/issues/5715

They don't seem to have recognised it, though...

PR #157, opened 2019-09-22
No out-of-window keys capturing, AltGr ≠ Ctrl (continued)

Cosmatevs:
The game should no longer capture keyboard keys while the window is out of focus. A lot/facade cannot be saved while an input is in focus.
AltGr is no longer treated as Ctrl… but, unfortunately, Ctrl isn't detected in `AltGr`+`LeftCtrl`. It's a makeshift solution, like #156.


TSOClient/tso.common/rendering/framework/io/InputManager.cs
```diff
@@ -54,6 +54,8 @@ public IFocusableUI GetFocus()
         /// <param name="keys"></param>
         public KeyboardInputResult ApplyKeyboardInput(StringBuilder m_SBuilder, UpdateState state, int cursorIndex, int cursorEndIndex, bool allowInput)
         {
+            if (!state.WindowFocused) { return null; }
```
riperiperi: I might need to change this for volcanic 2.0 (which adds keyboard states for winforms game views)

riperiperi:
Oops

PR #164, opened 2019-11-25
Implement Cheat Commands into VM

JDrocks450:
A command VMBudgetCheatCmd has been added for Simitone. It uses a VMCheatContext to define the cheat's behavior and parameters. Each cheat can have individually defined behavior or reuse behavior from another cheat. It also allows the cheat to see which avatar sent it when it is executed to allow in the future possible denial of access to certain cheats for individual players.


TSOClient/tso.common/utils/PathCaseTools.cs
```diff
@@ -12,7 +12,8 @@ public static class PathCaseTools
         public static string Insensitive(string file)
         {
             var dir = Directory.GetFiles(Path.GetDirectoryName(file));
-            return dir.FirstOrDefault(x => x.ToLowerInvariant().Replace('\\','/') == file.ToLowerInvariant().Replace('\\', '/'));
+            var name = Path.GetFileName(file).ToLowerInvariant();
```
riperiperi: Why is this changed? This will break linux and osx.

TSOClient/tso.content/framework/FAR1Provider.cs
```diff
@@ -189,29 +189,36 @@ public void Init()
...
+                                group = new List<Far1ProviderEntry<T>>();
+                                EntriesOfType[ext] = group;
+                            }
+                            group.Add(referenceItem);
                         }
-                        group.Add(referenceItem);
+                    }
+                    catch (Exception exception)
+                    {
+                        throw new Exception($"The FAR archive at: {archive} adding entry: {entry.Filename} threw the following exception: ", exception);
```
riperiperi: This doesn't really have anything to do with the PR. What could even throw in here?

TSOClient/tso.files/FAR1/FAR1Archive.cs
```diff
@@ -65,18 +65,26 @@ public FAR1Archive(string Path, bool v1b)
...
+                    Entry.DataOffset = m_Reader.ReadInt32();
+                    Entry.FilenameLength = (v1b) ? m_Reader.ReadInt16() : (short)m_Reader.ReadInt32();
+                    Entry.Filename = Encoding.ASCII.GetString(m_Reader.ReadBytes(Entry.FilenameLength));
+
+                    m_Entries.Add(Entry);
+                }
+                catch (Exception e)
+                {
+                    throw new Exception($"FAR1 Archive at: {m_Path} loading entry number: {i} threw an exception.", e);
+                }
```
riperiperi: I'm not sure anything could throw in here unless the files were corrupt, and it's outside the scope of this PR.

TSOClient/tso.simantics/NetPlay/Model/Commands/VMBudgetCheatCmd.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace FSO.SimAntics.NetPlay.Model.Commands
+{
+    /// <summary>
+    /// A command submitting a cheat with a typed parameter T
+    /// </summary>
+    public class VMBudgetCheatCmd : VMNetCommandBodyAbstract
```
riperiperi: This has no serialization so cannot be networked, but I suppose it's not a concern right now.

TSOClient/tso.common/utils/PathCaseTools.cs
JDrocks450: That line crashed my game on loading sounds. I believe it was due to my InstallPath variable in the registry having an extra slash at the end. I thought that other users might have a similar problem if their registry entry was the same way.

TSOClient/tso.simantics/Utils/VMCheatContext.cs
```diff
@@ -0,0 +1,97 @@
...
+                (bool success, int transferAmount, uint uid1, uint budget1, uint uid2, uint budget2) =>
+                {
+                        //check if we got the money? why should I? i mean cheating is bad after all
+                });
+            result = true;
+        }
+
+        private static void ExecuteMoveObjects(VM vm, VMAvatar caller, VMCheatContext context, out bool result)
+        {
+            vm.Context.TS1_EnableMod = context.Modifier;
```
riperiperi: Might be better having an object in Context to hold cheating state (move objects, draw routes) but that's a concern for later.

TSOClient/tso.simantics/VMContext.cs
```diff
@@ -54,6 +54,21 @@ public VMArchitecture Architecture
         public VMAmbientSound Ambience;
         public ulong RandomSeed;
 
+        /// <summary>
+        /// A simple work around for TS1 moveobjects cheat
+        /// </summary>
+        public bool TS1_EnableMod //to clean up changes see VMNetDeleteObjectCmd Verify and VMEntity IsUserMovable
```
riperiperi: Really don't like this - cheat state should be a child object to avoid cluttering VMContext, but I suppose we'll get there.

TSOClient/tso.common/utils/PathCaseTools.cs
riperiperi: Also it could actually break a ridiculous number of things since it will compare on filename and not the entire path. You should find out where the path is being joined, something is probably using + instead of Path.Join, or you can sanitize the path to not have a trailing slash when it is set on the content object. But not this.

JDrocks450:
Removed the unnecessary changes and moved cheat flag for moveobjects out of VMContext.

TSOClient/tso.simantics/entities/VMEntity.cs
```diff
@@ -1108,6 +1108,8 @@ private VMPlacementError IndividualUserMovable(VMContext context, bool deleting)
 
         public VMPlacementError IsUserMovable(VMContext context, bool deleting)
         {
+            if (Utils.VMCheatContext.CheatState.TS1_MoveObjects) // for ts1 moveobjects cheat
```
riperiperi: This shouldn't be a static variable. Ideally should be similar to other VM state classes, contained within the Context:
```
VMContext {
    VMClock Clock;
    VMArchitecture Architecture;
    VMCheatState Cheats;
}
```

riperiperi: Additionally, only cheats that can specficially apply to TS1 should be prefixed with TS1. MoveObjects could potentially apply to both.

TSOClient/tso.simantics/NetPlay/Model/Commands/VMBudgetCheatCmd.cs
riperiperi: While I'm suggesting the other change, this should just be VMNetCheatCmd, as it covers more than budget.

JDrocks450:
Added basic serialization to VMNetCheatCmd which can be expanded upon as cheats become more complex, renamed VMBudgetCheatCmd, and moved VMCheatState variable to where you suggested. Lastly, some VMCheatContext variables' types have been switched to smaller ones to keep serialized data sizes lower without affecting functionality.

riperiperi:
This looks much better, thank you for the changes!

JDrocks450:
Anytime! My bad for including all the extra nonsense :P

PR #166, opened 2019-12-06
various fixed of directory and file names (lower/upper case issue)

fHachenberg:
Trying to build FreeSO with Mono under Linux. Ran into some case inconsistencies (between build spec and actual directory name) for directory names.


riperiperi:
Thanks for this, I've been meaning to fix it for the longest time but basically never use linux. Not sure how it happened in the first place.

Issue #168, opened 2019-12-07
Build FreeSO with Mono under Linux


fHachenberg:
Hi there, thanks for your awesome work on FreeSO! 

I am trying to build under Linux using Mono msbuild. Apart from the minor naming issues already fixed in #166 & #167 , I still have some issue left.

Where can I find the following files? They don't seem to be present in mainline MonoGame either

    /FreeSO/TSOClient/FreeSO.sln.metaproj : error MSB3202: The project file "/FreeSO/TSOClient/../Other/libs/FSOMonoGame/MonoGame.Framework/MonoGame.Framework.WindowsGL.csproj" was not found. [/FreeSO/TSOClient/FreeSO.sln]
    /FreeSO/TSOClient/FreeSO.sln.metaproj : error MSB3202: The project file "/FreeSO/TSOClient/../Other/libs/FSOMonoGame/MonoGame.Framework/MonoGame.Framework.Windows.csproj" was not found. [/FreeSO/TSOClient/FreeSO.sln]
    /FreeSO/TSOClient/FreeSO.sln.metaproj : error MSB3202: The project file "/FreeSO/TSOClient/../Other/libs/FSOMonoGame/MonoGame.Framework/MonoGame.Framework.Linux.csproj" was not found. [/FreeSO/TSOClient/FreeSO.sln]
    /FreeSO/TSOClient/FreeSO.sln.metaproj : error MSB3202: The project file "/FreeSO/TSOClient/../Other/libs/FSOMonoGame/MonoGame.Framework/MonoGame.Framework.Net.WindowsGL.csproj" was not found. [/FreeSO/TSOClient/FreeSO.sln]
    /FreeSO/TSOClient/FreeSO.sln.metaproj : error MSB3202: The project file "/FreeSO/TSOClient/../Other/libs/FSOMonoGame/ThirdParty/Lidgren.Network/Lidgren.Network.WindowsGL.csproj" was not found. [/FreeSO/TSOClient/FreeSO.sln]
    /FreeSO/TSOClient/FreeSO.sln.metaproj : error MSB3202: The project file "/FreeSO/TSOClient/../Other/libs/FSOMonoGame/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.Windows.csproj" was not found. [/FreeSO/TSOClient/FreeSO.sln]

riperiperi:
[...]

PR #169, opened 2019-12-08
Fix dir cases 3

fHachenberg:
Found some more case issues after passing the hurdles described in #168. Sorry for the mass of PRs. :(

I guess that next some non-trivial issues with building on Linux will follow, so the next PRs will probably come from an ongoing branch for Linux.


riperiperi:
LGTM.

PR #170, opened 2019-12-12
Handle game being closed via events

dotequals:
With the Volcanic changes I noticed Volcanic's main window was still hanging open if the game was closed via the title bar close button. This PR adds the way of exiting the game the UI buttons use to anything that raises the OnExiting event which seems to solve the issue in my limited testing.


riperiperi:
What is this meant to achieve? `Game.OnExiting` is ONLY called when `Game.Exit()` is called - honestly surprised this does not cause an infinite loop, but I guess monogame devs have guarded against that.

`Process.GetCurrentProcess().Kill()` is not a valid way of closing volcanic as it's just a force close. Honestly don't know why it's still in GameFacade, but it's the only line that's active when you call here.

PR #172, opened 2019-12-16
Primitives Snap To Edges of other Primitives

JDrocks450:
When the setting is active, primitives will actively check the area around the mouse cursor for a primitive to snap the currently dragged or placing primitive to. The margins in between are designed to mimic the ones that already exist, but they can be changed for any reason easily (40px I think as of now). Placed primitives also will snap in the same way a dragged one will. 
(UIBuyMode change was to check something for MarshmellowBear)


riperiperi:
LGTM - In future though I'd just revert your changes for files that just add whitespace or whatever. I think it might be good to get it to snap to align with the connections (true/false), but those would be considerably more complicated. Additionally would probably like to save preferences to config or do a "hold shift" style toggle instead.

PR #173, opened 2019-12-22
Admin Red Chat Balloon

JDrocks450:
This will make the UIChatBalloon for an admin/moderator sim appear red instead of blue to stand out among users. It checks if the avatar's TSOState permissions attribute is set to Admin when the balloon text is updated. To maintain user customization settings, this does not have an impact on the chat text color of the balloon, but white is the easiest color to read in my opinion.


JDrocks450:
Second change is to fix pointer being drawn one pixel off from where it should be resulting in a minor graphical glitch.

riperiperi:
LGTM, I forgot that the colour was actually just a multiply. Nice catch on the 9-slice rounding error.

PR #174, opened 2019-12-23
Add Mentor Status to DB and PersonPage

JDrocks450:
This will make a user's mentor standing and moderator status visible on their person page. This uses two new custom ui badges that are designed to mimic the founder icon's style. Non-viewable changes are as follows:
- FSODataDefinition was edited to allow Avatar_ModerationStatus syncing between Client and Server, but the client will never attempt update ModerationStatus. Avatar_IsMentor was also added, and mentor status was added to the MyAvatar mask.
- Mentor status is updated by moderators in-game. An option was added to UIModMenu to toggle the current user's standing.
- Mentor status is User-Level so moderators only have to set mentor status on one of the mentor user's sims for convenience.
- In the db, the fso_mentors table stores the values by user_id and their current standing. 
- Avatar IsFounder status was not changed anywhere.


TSOClient/FSO.Server/DataService/Providers/ServerAvatarProvider.cs
```diff
@@ -333,6 +333,7 @@ private Avatar HydrateOne(DbAvatar dbAvatar, DbLot dbLot)
                 AvatarSkills_LockLv_Mechanical = dbAvatar.lock_mechanical
             };
             result.Avatar_PrivacyMode = dbAvatar.privacy_mode;
+            result.Avatar_IsMentor = DAFactory.Get().Users.GetMentorStatus(dbAvatar.user_id);
```
riperiperi: HydrateOne shouldn't call out to the DAFactory, it should be passed all relevant database objects. This makes it easier to batch them if we need to (like we do with properties).

TSOClient/FSO.Server.Database/DatabaseScripts/changes/0032_fso_mentors.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE IF NOT EXISTS `fso_mentors` (
```
riperiperi: What's the line of thinking of making this a separate table, rather than attaching it to the Users table? Are you thinking that more mentor related fields might be added to this table?

JDrocks450: Two things with that, number one was that I didn't necessarily know if you wanted the Users table to be appended with something like this at the time, knowing this now, I am agree this should be in the Users table. Number two is exactly that, I do think that badges could have a practical use in the future where a user can have a badge for something besides Mod/Admin, and Mentor. (Contributer?)

riperiperi: In that case, why not make it a "badges" table, with "badge_type" as a field? One avatar can have multiple badges - though for it to make the most general sense it might have to be linked on `avatar_id` rather than `user_id`, if you wanted to add badges for individual sims.

riperiperi: I could see such a system being useful in future if we decide to award badges for an event, and then show them on a badge tab in the Person Page.

PR #177, opened 2020-02-01
RackEOD: Bug fixes

TheArchitectFreeSO:
- When attempting to buy an outfit a Sim already owned, the transaction no longer silently fails _without_ charging the Sim's budget. An error message is displayed alerting the player.
- When attempting to buy an outfit in a category in which a Sim already owns 5 outfits, the transaction no longer silently fails _after_ charging a Sim's budget. An error message is displayed alerting the player and money is not deducted from the budget.

Rather than creating a new .cst entry for the "You cannot buy the same outfit twice" case, I just repeated the "You have no more room in your backpack" message.

### DIFF
--- a/TSOClient/tso.client/UI/Panels/EODs/UIRackEOD.cs
+++ b/TSOClient/tso.client/UI/Panels/EODs/UIRackEOD.cs
@@ -35,21 +35,34 @@ namespace FSO.Client.UI.Panels.EODs
 
         private void PurchaseErrorHandler(string name, byte[] type)
         {
-            // I already own this outfit
+            string title = GameFacade.Strings.GetString("264", "5"); // "Purchase Item"
+            String message = "";
+            // I already own this outfit - should do a custom error message but that means a new .cst entry...
             if (type[0] == 0)
             {
-
+                message = GameFacade.Strings.GetString("264", "10"); // "You have no more room in your Backpack"
             }
             else if (type[0] == 1)
             // I already own 5 outfits of this type
             {
-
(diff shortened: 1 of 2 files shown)


TheArchitectFreeSO:
error msg

TSOClient/tso.client/UI/Panels/EODs/UIRackEOD.cs
```diff
@@ -27,6 +27,44 @@ protected override void InitUI()
...
+        {
+            base.InitEOD();
+            BinaryHandlers["rack_buy_error"] = PurchaseErrorHandler;
+        }
+
+        private void PurchaseErrorHandler(string name, byte[] type)
+        {
+            string title = GameFacade.Strings.GetString("264", "5"); // "Purchase Item"
+            String message = "";
+            // I already own this outfit - should do a custom error message but that means a new .cst entry...
```
riperiperi: If they didn't have this error originally, I guess you could own multiple of the same? Not that I'd want to change it to that.

riperiperi:
This is definitely an improvement. There's still a chance for a race... ...ok maybe never - you'd have to obtain clothes from two places at the same time and that is not very probable.

The way to deal with the race condition, if it were important, would be to catch when the outfit fails to add and refund the money. But this is definitely 99.99% of the way there and I'm all for that.

TSOClient/tso.client/UI/Panels/EODs/UIRackEOD.cs
TheArchitectFreeSO: Yeah, exactl. He only had a client side validation to enable the button. Circumventing that would have failed on outfits delivery, too.

PR #181, opened 2020-06-09
Random skintone in Body Strings

andrewknoll:
The field for **skin tone** (field 14) in the **body strings** for an NPC can hold now **more than one value**.
The syntax, for such field is now the following:
- **field14** = **S**
- **S** = **skt** **sc** **S***
- **skt** = 'lgt' | 'med' | 'drk'
- **sc** =(';')+

The skin tone of the NPC will be determined **randomly** amongst all the options given in the sequence.
Notice:
- The sequence cannot be empty.
- The sequence can have a single item (therefore compatible with pre-existing NPCs which weren't modified in this PR).
- The sequence can have repeated elements (it would only increase the chance of drawing the skin tone of such element).
- Extra semicolons will be ignored.
- Using a skin tone other than those defined by the syntax rule "skt" may have unexpected results.

The following NPCs should spawn with random skin tones:

- Pizza delivery men/women
- Maid
- Repairman
- Gardener
- Restaurant Patrons (all levels)
- Restaurant Busboys / busgirls (all levels)
- Restaurant Chefs (all levels)
- Nightclub DJs (all levels)
- Nightclub Platform Dancers (all levels)
- Nightclub Dancefloor Dancers (all levels)

PR #183, opened 2020-06-23
Curtains/blinds available to put on windows

Cosmatevs:
As a side effect they become layerable, just like hanging flowers. I think it's a fair enough solution anyway. These changes also affect existing curtains and blinds.

Should I also add something to *TSOClient/tso.content/FSO.Content.csproj* like Andrewkoll did in #181? I'm asking because of this line:
` <Content Include="Content\Patch\**\*.*">`


riperiperi:
It's not necessary for patches. Thanks for the PR, I'll have a look later today.

riperiperi:
That wasn't later today, but I did check it out and it all looked good. Even added the Load tree in there, nice.

The exclusive placement flag rules might change in the future once I have time to figure them out (like in a way that doesn't let you put cordless phones over doors like ts1 unleashed), but for now this is golden.

PR #189, opened 2021-01-04
GameshowBuzzerEOD

TheArchitectFreeSO:
While we need to test this in closed beta, the GameshowBuzzerEOD was made possible thanks to the efforts of @andrewknoll and his tireless efforts to create three new objects: The Host Buzzer Podium, the Standing Player Podium, and the Seated Player Podium.

The two player podiums display the score of the connected player, while the host merely animates the charming connected host avatar.

Each podium type, host and player, use their own plugin handlers that interact with each other where necessary. Any type and amount of player podium objects can be placed in a lot and (as long a host is connected to a host podium) the host plugin will detect connected players, regardless of their location in the lot. Multiple host podiums can also exist, which use branding methods to avoid overlapping/sharing players—though a thorough test of such advanced setups will be a focus of the closed beta test. I can only effectively test 6 Sims at once, after all .

All graphical work and animation sequences are thanks to @andrewknoll and they add to the magic of a live gameshow. This will be the first EOD to use the new Extra_tall panel that was added to LiveMode.cs a long time ago.


TheArchitectFreeSO:
Additionally
•A minor error was fixed in RouletteEOD to avoid throwing a null exception error when clicking in the UIEOD.
•A minor error was fixed in MazeEOD to avoid a very rare null exception error for the logic player.
•There may be one final minor fix to the logic of HoldemCasinoEOD to slightly increase the minimum required to stock a table to open it, due to a calculation error.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODGameshowBuzzerHostPlugin.cs
```diff
@@ -0,0 +1,927 @@
...
+            for (int index = 0; index < 4; index++)
+            {
+                if (index == playerIndexToSkip)
+                    continue;
+                var player = ConnectedContestants[index];
+                if (player != null)
+                {
+                    if (player.EODType.Equals(VMEODGameshowBuzzerPluginType.Player))
+                    {
+                        if (!onlyActiveBuzzers || onlyActiveBuzzers && player.MyBuzzerEnabled)
```
riperiperi, line 661: This is a rather confusing statement. I think `!onlyActiveBuzzers || player.MyBuzzerEnabled` would function identically - it is true if non-active buzzers are allowed, OR if the player's buzzer is enabled.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODGameshowBuzzerHostPlugin.cs
```diff
@@ -0,0 +1,927 @@
...
+                            players.Add(player);
+                    }
+                }
+            }
+            return players;
+        }
+        private int GetPlayerIndex(VMEODGameshowBuzzerPlayerPlugin player)
+        {
+            int playerIndex = -1;
+            lock (PlayerConfigLock)
```
riperiperi, line 671: Nested locks always give me super anxiety. If you do this, make double sure that they are always locked in the same order, otherwise you can encounter a deadlock.

TSOClient/tso.simantics/NetPlay/EODs/Handlers/VMEODGameshowBuzzerPlayerPlugin.cs
```diff
@@ -0,0 +1,220 @@
...
+        {
+            string evt = "Buzzer_Player_Other_Correct";
+            if (were_they_correct == 0)
+                evt = "Buzzer_Player_Other_Incorrect";
+            // execute client event to update UI for other's answer
+            Send(evt, opponentName);
+            ReactToOtherJudgment(were_they_correct);
+        }
+        /// <summary>
+        /// Another player answered a question and the host indicated if they were correct or incorrect
```
riperiperi, line 103: This xml doc is completely incorrect.

TSOClient/tso.content/FSO.Content.csproj
```diff
@@ -176,6 +176,19 @@
     <Content Include="Content\Textures\EOD\Band\bandskilllevels.png">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="Content\Textures\EOD\Buzzer\eod_lightsframe1.png" />
```
riperiperi, line 179: This content must be copied to the output directory.

riperiperi: (and all others below it)

riperiperi (approved):
Very impressive stuff, just a few comments and notes. This is clearly a really complicated EOD and people will have a lot of fun with it. I really appreciate all the configuration options the host has - since they will be coming up with their own game rules obviously. (when are we recreating pointless in freeso?)

- There should be an empty line between functions in a class.
- You use reflection to generically set properties with one method... Which is fine, but it's preferable if it's only used as a last resort.
- I really don't like all this locking stuff, especially because deadlocking lot threads in a server that runs around 30-40 properties at once can cause issues, as the property then cannot be closed. I was thinking of potentially moving the events that can happen in another thread to happen in a callback queue (such as disconnect), so that you can do everything on the simantics thread and not have to worry about thread safety.
- All EODs do this right now, but using System.Timer is not appropriate within a game. We have internal functions for timeout and interval timers, based on the frame count, with callbacks that happen in the game thread. This would also prevent any weird threading issues caused by timer callbacks.

**There is one important change needed:**
- The new content must copy to the output directory.

riperiperi:
I will make necessary changes myself, since time is pretty short. Thanks a lot for the contribution!

TSOClient/tso.content/FSO.Content.csproj
TheArchitectFreeSO: I had some weird VS issues with this—it wanted to make 10000 changes by shifting everything down unnecessarily...so I must've messed up doing some line-item changes. 

Thanks for making these changes since time is more challenging now 👶

PR #194, opened 2021-05-10
Chat color changes

andrewknoll:
- Visitors get a brand new green chat color that matches the icons on the property sim list.
- NPCs get a brand new gray chat color that matches the icons on the property sim list.
This is to make the roommates more easily recognizable, especially for newbies.


andrewknoll:
Not sure why all those commits appeared when I merged... sorry!!

riperiperi (approved):
LGTM, nice touch!

PR #196, opened 2021-06-02
Random splash screen

andrewknoll:
- If there is a directory called "Content/SplashScreens", the game will use a random PNG image from that directory.
- The file "Content/setup.png" still works as an alternative should this directory not exist or be empty.


riperiperi (approved):
LGTM. I might change this a bit to counter the fact that the updater cannot delete files right now.

PR #198, opened 2021-07-04
Handle DirectoryNotFoundException if the SplashScreens directory does not exist

inb40:
I recently started working with the repo and saw that this throws an Exception on a clean repo install, apparently this folder is optional but its DirectoryException is not handled.


riperiperi:
Nice catch - forgot this type of exception would be thrown.

riperiperi (approved):

Issue #83, opened 2017-11-03
Username and password field cannot be selected with shift+end


Invictaz:
Username and password fields cannot be selected before deletion with shift+end. Only allows delete or backspace keys to clear the field. 

Request is to have this available so fields can be cleared more easily.

![image](https://user-images.githubusercontent.com/20338545/32398640-d4f5a3f6-c0f0-11e7-8de1-d18c4a23be78.png)

PR #201, opened 2021-07-09
#83 - adds shift+end and shift+home functionality

ssinakhot:
### What was changed?

- Adds the shift+end and shift+home functionality to the InputManager code
- Adds the end and home functionality to bring cursor to beginning or end
- Removed extra tabs from the ending portion of the long if-else statement to make it match up

### How to test?

- Type something in a text box and move the cursor to the center
- Hit Shift+End to select everything from cursor to the end
- Hit Shift+Home to select everything from start to cursor
- Use End to move cursor to end
- Use Home to move cursor to beginning

### DIFF
--- a/TSOClient/tso.common/Rendering/Framework/IO/InputManager.cs
+++ b/TSOClient/tso.common/Rendering/Framework/IO/InputManager.cs
@@ -271,12 +271,23 @@ namespace FSO.Common.Rendering.Framework.IO
                                     didChange = true;
                                 }
                                 break;
-                            }
-                        } else {
-                            result.UnhandledKeys.Add(key);
-                            processChar = true;
                         }
+                    } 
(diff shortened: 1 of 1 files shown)


riperiperi:
A few issues, but I'm glad it's being added:

- This breaks selecting text with shift and the arrow keys.
  - The other navigation keys go through `UnhandledKeys` to the `UITextEdit`. Check out `UITextEdit:504` for the location where these keys are handled. It might be better to handle the new keys here, too, or at least allow the behaviour to follow through to here if shift is pressed with any other key.
- Copy/Paste doesn't work correctly when selecting using shift+home, and selecting with non-shift home and end does not revert to a . Be careful what you're doing with the selection start and end. Start must be before end, and a cursor with no selected characters should have the selection end as -1.

PR #203, opened 2021-07-09
QoL - add flashing to interactions that require responses

ssinakhot:
https://cdn.discordapp.com/attachments/193181767585366017/863057422322237500/interaction.gif

### What Changed?

- Added flashing of opacity to interactions to make them more noticeable when they need a response

### How to test?

- Get a 2nd person to send an interaction that requires a response


ssinakhot:
Flashing the active border will be more "on brand"

PR #204, opened 2021-07-10
Possible? Fix for when you click into the Window without focus and it doesn't properly select UIElements

ssinakhot:
What is happening is that when the game is not focused and you go to click on a UIElement, the game gets focused and doesn't trigger the UIElement click properly. Rather it passes the click down to the lot/terrain world instead. 

You can see this happen when you try to click into the chat window. 

Note: I am not 100% sure what this change can also affect. I have not seen any negative affects on the Windows FreeSO of things. So don't take this merge without proper review.

### What changed?

- Allow mouse events to be handled when window is not focused.

### How to test?
- Launch game & click on a UIElement without having Window Focus
- It should properly click and trigger hover over tooltips instead of being treated like a lot click

### DIFF
--- a/TSOClient/tso.common/Rendering/Framework/GameScreen.cs
+++ b/TSOClient/tso.common/Rendering/Framework/GameScreen.cs
@@ -108,16 +108,8 @@ namespace FSO.Common.Rendering.Framework
             else
             {
                 //single mouse state
-                if (hasFocus)
-                {
-                    State.MouseState = Mouse.GetState();
-                    State.KeyboardState = Keyboard.GetState();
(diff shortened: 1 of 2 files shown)


riperiperi:
This change allows the mouse cursor to act through windows, which won't be great when alt-tabbed. I'm not sure what causes the specific behaviour causing the lot click to go through, it's worth investigating that more. (perhaps the mouse position tested when clicking from unfocused is (0, 0), so it goes through the elements? idk)

https://gyazo.com/69c5f11aee90858f256d48f6cb8e0a66

PR #206, opened 2021-07-17
fix: pie menu no longer glitches for 1 frame.

ddfczm:
Pie menu no longer displays 1 frame of the background at full size before resizing. The Image component considers Width/Height==0 to mean don't scale the image at all (has no way to represent unset/null). Instead of changing the image component and risking breaking other usages this just starts the background at 2x2.


riperiperi (approved):
Nice catch. Don't know how I didn't notice this one in... years, but when I check it is indeed a problem.

PR #207, opened 2021-07-17
fix: person grid navigation buttons now disable/enable at the start/end

ddfczm:


TSOClient/tso.client/UI/Panels/UIPersonGrid.cs
```diff
@@ -101,5 +118,13 @@ public void PreviousPage()
             Page--;
             DrawPage();
         }
+
+        public int NumPages
+        {
+            get
+            {
+                return Math.Max(1, (int)Math.Ceiling((double)Display.Count / (double)(Columns * Rows)));
```
riperiperi, line 126: `((Display.Count - 1) / (Columns * Rows)) + 1`
Should give a number of pages without conversion to and from double. This should also give "1" page for 0 items.

ddfczm: 👍 updated

riperiperi (approved):
Looking great, thanks!

PR #243, opened 2022-07-25
[VM] 3D audio panning

LazyDuchess:
Adds specific code for panning entity sounds while in 3D mode, as previously the 2D code was used causing it to jump around and be inaccurate.
Avatars are also now added to the VM sound entities when they play a sound from an event, so that it pans around as you move the camera rather than staying in place.


riperiperi (commented):
Looks good, though I will need to wait until to tomorrow to test it.

riperiperi:
Gave it a test, sounds great. Nice work!

riperiperi (approved):

riperiperi:
Your unrelated change also looks good. 🥴

LazyDuchess:
sorry bout that ♥♥♥♥

PR #258, opened 2022-11-11
Disabled flags TSO Generic calls.

andrewknoll:
+ Adds FSOIsStackObjectPendingRoommateDeletion: Returns true when the Stack Object is pending deletion after their owner moved out of the property.
+ Adds FSOIsStackObjectAllowedByLotCategory: Returns true if the Stack Object is allowed in the current lot category.


riperiperi (approved):
LGTM, I can see how these might be useful.

PR #260, opened 2022-11-20
New FSOCatalogName wildcard for string resources

andrewknoll:
Using `$Object` or `$DynamicObjectName:` in a string resource returns the name of the object. In case of a multi-tile object, this name will be the name of such tile, and not the master object. See example:
image

Leading to a rather unnatural way of naming things for the end user, or even to misleading information:
image

This PR introduces `$FSOCatalogName`, which takes the master definition (if applies) of the object and returns the name as shown in their catalog strings.

Note 1: `$CatalogLocal:` always returns the catalog description, rather than the name.
Note 2: Needs to patch the Salvage interaction and [REDACTED].


andrewknoll:
Fixed as in 8ed8e5cc1a9b0f68905b54d1b08b35b0493460cf

PR #261, opened 2022-11-29
"Never Clamp" relationship primitive flag

andrewknoll:
This flag allows for a relationship to go over (and below) the limits of [-100, 100] for relationships, useful for new ways of interacting with objects, and one that is to be released soon.

### DIFF
--- a/TSOClient/tso.simantics/Primitives/VMRelationship.cs
+++ b/TSOClient/tso.simantics/Primitives/VMRelationship.cs
@@ -275,6 +275,7 @@ namespace FSO.SimAntics.Primitives
                 else Flags &= unchecked((byte)~128);
             }
         }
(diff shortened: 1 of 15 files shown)


TSOClient/tso.simantics/Primitives/VMRelationship.cs
```diff
@@ -275,6 +275,15 @@ public bool FSONeverPersist
                 else Flags &= unchecked((byte)~128);
             }
         }
+        public bool FSONeverClamp
```
riperiperi, line 278: Newline before this.

riperiperi (commented):
Thanks for this. I double checked the flags to maks sure `64` isn't taken and it looks like you're good.

PR #266, opened 2022-12-03
Fixes for DJ/Dancer Job Outfits and Fixes for All Outfits for all jobs.

thertzelle:
### What
- Fixing Outfits for DJ and Dancer Job per [TSOMania](https://www.tsomania.net/gameguides/jobs_dj_levels.php)
- Fixing equation for determining 'level' which is what is used for determining which array element of a job outfit to select, see pull request for tabled fix data points and what this changes.
- Fixed a MIN value situation, where the value should never become 3, because 3 would be index out of bounds.

Below I have put together a table, that represents the current values, and the goal values. Job Grade is the current int of job. Expected result column built based on TSOMania game guide, the current computed C# value without round, and the tested C# value with round (the fix), and finally just a confirmation column, on observing the current values using Volcanic.

|   JOB GRADE  |   EXPECTED RESULT  |   C# Current - (jobLevel - 1) / 4;  |   C# With New Code  |   Current (Via Volcanic)  |
|--------------|--------------------|-------------------------------------|--------------------------------------------------|---------------------------|
|   0          |   0                |   0                                 |   0                                              |   0                       |
|   1          |   0                |   0                                 |   0                                              |   0                       |
|   2          |   0                |   0                                 |   0                                              |   0                       |
|   3          |   1                |   0 (wrong)                                |   1                                         |   0                       |
|   4          |   1                |   0 (wrong)                                |   1                                         |   0                       |
|   5          |   1                |   1                                 |   1                                              |   1                       |
|   6          |   1                |   1                                 |   1                                              |   1                       |
|   7          |   2                |   1  (wrong)                               |   2                                            |   1                       |
|   8          |   2                |   1  (wrong)                               |   2                                          |   1                       |
|   9          |   2                |   2                                 |   2                                              |   2                       |
|   10         |   2                |   2                                 |   2                                              |   2                       |

### DIFF
--- a/TSOClient/tso.simantics/Engine/VMSuitProvider.cs
+++ b/TSOClient/tso.simantics/Engine/VMSuitProvider.cs
@@ -39,14 +39,14 @@ namespace FSO.SimAntics.Engine
 
                 new long[] { //dj
                     0x5900000000D,
-                    0x5920000000D,
                     0x5940000000D,
+                    0x5920000000D,
                 },
 
                 new long[] { //dancer
                     0x5910000000D,
-                    0x5930000000D,
                     0x5950000000D,
+                    0x5930000000D,
                 }
             },
             new long[][] { //female
@@ -70,14 +70,14 @@ namespace FSO.SimAntics.Engine
 
                 new long[] { //dj
                     0x5810000000D,
-                    0x5830000000D,
                     0x5850000000D,
+                    0x5830000000D,
                 },
 
                 new long[] { //dancer
                     0x5820000000D,
-                    0x5840000000D,
                     0x5860000000D,
+                    0x5840000000D,
                 }
             }
         };
@@ -222,7 +222,7 @@ namespace FSO.SimAntics.Engine
                             if (context.VM.TS1) return null;
                             var job = avatar.GetPersonData(VMPersonDataVariable.OnlineJobID);
                             if (job < 1 || job > 5) return null;
-                            var level = Math.Max(0, Math.Min(3, (avatar.GetPersonData(VMPersonDataVariable.OnlineJobGrade) - 1) / 4));
+                            var level = Math.Max(0, Math.Min(2, Math.Round((avatar.GetPersonData(VMPersonDataVariable.OnlineJobGrade) - 1) / 4.0)));
                             return (ulong)(JobOutfits[male?0:1][job-1][level]);
                         case VMPersonSuits.DefaultSleepwear:
                             return avatar.DefaultSuits.Sleepwear.ID;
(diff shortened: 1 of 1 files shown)


thertzelle:
This seemed like it was merged in update L, but then taken back out of update M? @riperiperi

riperiperi:
This was never merged.

TSOClient/tso.simantics/Engine/VMSuitProvider.cs
```diff
@@ -222,7 +222,7 @@ public static object GetSuit(VMStackFrame context, VMSuitScope scope, ushort id)
                             if (context.VM.TS1) return null;
                             var job = avatar.GetPersonData(VMPersonDataVariable.OnlineJobID);
                             if (job < 1 || job > 5) return null;
-                            var level = Math.Max(0, Math.Min(3, (avatar.GetPersonData(VMPersonDataVariable.OnlineJobGrade) - 1) / 4));
+                            var level = Math.Max(0, Math.Min(2, Convert.ToInt32(Math.Round((avatar.GetPersonData(VMPersonDataVariable.OnlineJobGrade) - 1) / 4d))));
```
riperiperi, line 225: What is this meant to do? I've got a few complaints.

`Convert.ToInt32` shouldn't be used, cast with `(int)`.
`4d` instead of `4.0` for double literals is weird. Also, avoid using doubles for integer math.

If you want the integer division to round normally, add (divisor / 2) to the value before dividing it. In this case, that would change `- 1` to `+ 1`.

riperiperi:
```suggestion
                            var level = Math.Max(0, Math.Min(2, (avatar.GetPersonData(VMPersonDataVariable.OnlineJobGrade) + 1) / 4));
```
Maybe this is what you want?

thertzelle:
I'll take a look at the conversion change in a little bit. Strange that everyone reported that shortly after I submitted this PR, the outfits were 'fixed', but after the M update they reverted back to the previous invalid state. I'll update the PR. Thanks for taking a look.

riperiperi:
Yeah, it looks like it was an issue with the update builder. It took the azure build from this PR instead of the main branch... should be sorted now. I don't think this happened with any other PR (that wasn't an inconsequential dependabot one).

thertzelle:
`var level = Math.Max(0, Math.Min(2, ((int)avatar.GetPersonData(VMPersonDataVariable.OnlineJobGrade) + 1) / 4));`
I updated the code and here is the results:

```
Job Grade: 0 Level: 0

Job Grade: 1 Level: 0

Job Grade: 2 Level: 0

Job Grade: 3 Level: 1

Job Grade: 4 Level: 1

Job Grade: 5 Level: 1

Job Grade: 6 Level: 1

Job Grade: 7 Level: 2

Job Grade: 8 Level: 2

Job Grade: 9 Level: 2

Job Grade: 10 Level: 2

PR #268, opened 2023-01-20
Enable nightclub03_00 lot file based on jobType.

thertzelle:
### What
- For DJ / Dancer the `jobType` variable here is always 2 (it would be bad if it was greater than 2 because JobGradeToLotGroup[jobType] of anything beyond 2 isnt defined).
- So we can safely change this check to `jobType == 2` to ensure that DJ and Dancer jobs will have the opportunity to generate a random number either 1 or 2, instead of just 1.


thertzelle:
@riperiperi Sorry to at you in a message, but this repo isn't configured to let me add reviewers.

Please take a look when you get a moment.

riperiperi (approved):

Issue #215, opened 2022-01-05
Give the option for property owners/roommates and Administrators to reset money objects to their default state.


tomgrice:
Another player suggested that it would be helpful if property owners/roommates were able to easily reset skill money objects (e.g. potions table, gnome workbench) to their default state.

Would be useful in situations where a low skill player intentionally or unintentionally starts work on a table and logs out/leaves the property. Currently if a low skill level sim starts working at a table and a higher skill level sim completes it, the pay-out will be a low amount. This would solve situations where people troll by starting a large amount of skill objects with no intention of finishing them.

I'm putting this up as a suggestion, feel free to shoot it down if needed, but if it seems a good idea I might have a look and have a go at implementing when I get time if nobody else picks it up.

PR #272, opened 2023-08-17
Add interactions to reset all solo money objects

LazyDuchess:
Adds a new non-immediate interaction to each solo money object to reset their state, accessible only by property owners and roommates. 

Idea brought up in #215, might be niche but people running money lots could find it useful.

https://github.com/riperiperi/FreeSO/assets/42678262/aad67650-5da2-43ce-bcd8-dff829767294


riperiperi:
Sorry for the delay. To avoid duplicating the dialog across all objects, dialog strings for clearing these could be in global `skillobjects.iff`, and maybe part of the tree?

PR #275, opened 2023-12-11
Robot Factory Pay Raise?

Akzelius:
The robot factory job is less popular than the restaurant because on average it pays less, and an obvious solution to this is to improve the scaling of the robot payouts to make it a more viable alternative while not going overboard since they are two entirely different types of activities after all.

Diner payouts obviously vary a lot but on average when working with 2 people my experience is that you can earn in tips around $500 to $1000 on the levels 0-2, around $2000 to $3000 on the levels 3-6 and around $3500 to $5000 on the levels 7-10.

In the robot factory when working with two people you usually make anywhere between 10 and 20 robots so let's say 15 as an average to compare. This means with the current payouts on levels 0-2 you make between $450 and $750, on levels 3-6 you make between $900 and $1350, and on levels 7-10 between $1500 and $1950.

As can be seen from this the difference is not that big on the lower levels but on the higher levels you make half of what your restaurant equivalent does. My proposal to make the jobs a little more equal so that more people will choose to work the robots is to keep the baseline at $30 per robot but increase the pay raise between promotions from +10 to +20, resulting in the following changes:

Level 0: $30/robot
Level 1: $50/robot
Level 2: $70/robot
Level 3: $90/robot
Level 4: $110/robot
Level 5: $130/robot
Level 6: $150/robot
Level 7: $170/robot
Level 8: $190/robot
Level 9: $210/robot
Level 10: $230/robot

Working with the same comparison as before this would mean that with a 15 robot average you would make on levels 0-2 between $450 and $1050, on levels 3-6 between $1350 and $2250, and on levels 7-10 between $2550 and $3450. This should make the gap between factory and restaurant less noticeable. If it still sounds low keep in mind that a good factory player on a round without a disastrous amount of repairs usually does 15 as a minimum rather than an average.

### DIFF
--- a/TSOClient/tso.content/Content/Patch/Tuning/oj-robotfactorycontroller.otf
+++ b/TSOClient/tso.content/Content/Patch/Tuning/oj-robotfactorycontroller.otf
@@ -7,6 +7,10 @@
       (10/15/20/25/25/30/30/35/40/45/50)
     - increased roughly 2.5x, gradual scaling
       (30/40/50/60/70/80/90/100/110/120/130)
+
+   UPDATED CHANGES
+   - same baseline but higher pay raise per promotion to be a more viable alternative to restaurant job
+      (30/50/70/90/110/130/150/170/190/210/230)
    -->
 
    <T i="4096" n="Tuning - grade 0">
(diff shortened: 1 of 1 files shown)


riperiperi:
About making balance changes like this. I don't know who you are ingame or in discord, so I don't know if I can trust you. I do agree that the robot job is somewhat underpowered, but that might also mean changes in competing job's pay might be needed too.

If someone is able to validate the earnings for both restaurant and robot, for _both_ a multiplayer session and solo, then it can be considered.

PR #276, opened 2023-12-20
Improve extreme color calculation for DXT compress

kaetemi:
Take average color, then use vector math to get the furthest color from the average, then again find the furthest color from the furthest color.

### DIFF
--- a/TSOClient/tso.common/Utils/TextureUtils.cs
+++ b/TSOClient/tso.common/Utils/TextureUtils.cs
@@ -802,7 +802,7 @@ namespace FSO.Common.Utils
 
 			// Alpha is determined in RGB565 representation
 			// If alpha, Color 1 is greater or equal to color 0
-            // If no alpha, Color 0 is greater or equal to color 1
+            // If no alpha, Color 0 is greater than color 1
 			if (hasAlpha0 != (leftBin < rightBin))
             {
 				// hasAlpha0 && (leftBin >= rightbin)
--- a/TSOClient/tso.common/Utils/TextureUtils.cs
+++ b/TSOClient/tso.common/Utils/TextureUtils.cs
@@ -535,15 +535,15 @@ namespace FSO.Common.Utils
                         }
                     }
 
-                    Color minCol, maxCol;
-                    GetMinMaxColor(block, out minCol, out maxCol);
+                    byte minAlpha, maxAlpha;
+                    GetMinMaxAlpha(block, out minAlpha, out maxAlpha);
 
-                    //emit alpha data
+					//emit alpha data
 
-                    result[blockI++] = maxCol.A;
-                    result[blockI++] = minCol.A;
+					result[blockI++] = minAlpha; // VERIFY: Check if this is reverted
+                    result[blockI++] = maxAlpha;
 
-                    var alpha = GetAlphaIndices(block, minCol, maxCol);
+                    var alpha = GetAlphaIndices(block, minAlpha, maxAlpha);
 
                     result[blockI++] = (byte)((alpha[0] >> 0) | (alpha[1] << 3) | (alpha[2] << 6));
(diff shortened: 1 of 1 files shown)


kaetemi:
Both DXT1a and DXT5 tested and working.

Before:
image

After:
image

kaetemi:
Slightly better with weighted colors for distance calculation.

image

riperiperi:
Thanks for having a look at this issue. Two quick notes before I give it a review later (not at a pc for a while):

- Indentation looks incorrect in the diff view, it should be 4 spaces on this project instead of tabs.
- This code is currently performance critical when the texture recompression is enabled, as it happens on the render thread the first time the texture is accessed. I don’t want to make it any slower, so ideally it shouldn’t exhaustively search for endpoints if it doesn’t already. I wrote this ages ago, so I can’t remember its current complexity.
  - I’d be willing to trade off performance if this were done asynchronously or had some kind of cache.

kaetemi:
Previous code does one iteration and checks for both minimum and maximum in that pass in each color individually. (Which leads to completely incorrect color mixes when the endpoints are different colors. E.g. green to blue becomes black to cyan.)

New code does one iteration to find the average, one iteration to find the furthest, and another iteration to find the opposing endpoint.

Both then iterate again to find which of the 4 colors in the endpoint gradient is nearest to each pixel.

Agree a cache would be good eventually. Haven't noticed any performance drop, at least, and looks better. Could put a performance timer on it to verify.

kaetemi:
Somehow faster on my machine. Ryzen 5 2600.

> new code:
> Total time: 598ms
> Total blocks: 748286
> Time per million blocks: 799ms
> 
> old code:
> Total time: 712ms
> Total blocks: 747132
> Time per million blocks: 952ms
> 
> 
> New code (second run):
> Total time: 641ms
> Total blocks: 746612
> Time per million blocks: 858ms
> 
> Old code (second run):
> Total time: 780ms
> Total blocks: 754988
> Time per million blocks: 1033ms
> 
> 
> New code (third run):
> Total time: 747ms
> Total blocks: 773352
> Time per million blocks: 965ms
> 
> Old code (third run):
> Total time: 851ms
> Total blocks: 778296
> Time per million blocks: 1093ms

TSOClient/tso.common/Utils/TextureUtils.cs
```diff
@@ -659,111 +662,166 @@ private static byte[] GetAlphaIndices(Color[] block, Color minCol, Color maxCol)
...
+                ? new Color(0, 0, 0)
+                : new Color(r / t, g / t, b / t);
 
-                if (col.R > maxCol.R) maxCol.R = col.R;
-                if (col.G > maxCol.G) maxCol.G = col.G;
-                if (col.B > maxCol.B) maxCol.B = col.B;
+            // Find color furthest from average
+            int leftDist = 0;
+            int leftIdx = 0;
+            for (int i = 0; i < 16; i++)
```
riperiperi, line 769: This loop can be combined with the next one. That would skip duplicating the loop overhead, and let you do the fetch from block[i] into local variable once instead of (at least) twice.

riperiperi (approved):
Sorry for the delay, I wasn't near a PC for a few weeks.

Results are looking a lot better for me, and after looking at the code the only additional cost is one additional loop for the left/right selection after getting the average (though currently two).

Changes to GetColorIndices and GetA1ColorIndices likely improved performance, as you've removed heap allocations for the Color[]s, unrolled loops and saved a distance calculation. There's slightly more complex distance calculation but it's totally worth it.

All around great work. I can merge without combining the loops if you don't have the time, just let me know.

kaetemi:
Alright, sounds good.

I don't believe the loops can be combined? The loop depends on the outcome of the previous loop.

riperiperi:
Ah, I misread it - thought left col wasn't used. Looks good to me then.

PR #280, opened 2024-12-22
Correct capitalization for the Protobuild command in documentation

SegerEnd:
This PR updates the capitalization of the Protobuild.exe command in the documentation, making it easier to copy and paste directly instead of typing it myself from the screenshot :)


riperiperi (approved):
I copied this from the azure config, it's not a problem there as the build runs on Windows... Makes sense that it would fail on linux, though.

PR #283, opened 2025-09-20
Upgrade FreeSO to .NET 9.0

SegerEnd:
Hi! I’m upgrading FreeSO to .NET 9.0, It is currently working and playable! There are still some areas for improvement. I wanted it to place it here already so it can be used, and maybe collaborate together.

##
- **Graphics**: Now using the latest `MonoGame` version.
- **Dependencies**: All major dependencies are now managed via NuGet. I updated them to the latest versions where possible, though some are still tied to .NET Framework 4.5.
- **Build process**: Protobuild is no longer required you can clone, build, and run directly. (hope it works on other machines as well haha )
- **Performance & optimization**: .NET 9.0 may provide runtime performance improvements, and the code can be further optimized to fully leverage the new framework.

##
- I wanted the project to have AOT Compilation, that currently fails because **Ninject** and some other are still .NET 4.5 packages and some rely on reflection.
  - Should we replace Ninject with `Microsoft.Extensions.DependencyInjection` and consider modern alternatives for most older libraries?
  - Could the project structure also be improved while updating these libraries?

##
- Makes it easier for new contributors to build and run the project.
- Removes the need for Protobuild, simplifying the setup.
- Updates the codebase to current .NET and MonoGame standards.
- Opens opportunities for better performance and further optimization.


riperiperi:
Awesome, definitely didn't expect someone to attempt this before I've completed the archive client, and base on top of it, too. I'll have a look through specific changes as soon as I can, and run it locally to assess how more complicated things work. (like FSO.IDE, which needs to be included despite being based on Windows.Forms)

The more work I've been putting into the archive client, the more I've been tempted to port it to .NET 9 to get the newer features and performance improvements (why should anyone need to avoid Enum.HasFlag? and other similar quirks... then obviously spans, intrinsics, a lot more). I do think modernizing everything is worth trading off the older Windows support.

A few preliminary notes:
- I'm making a lot of core game changes in the Archive client because I'm adding some new game features to improve the exploration experience.  This could result in some big merge conflicts in the future.
  - That does make it good to get these changes in ASAP... So it might not be important to get things like AOT support working for this PR.
  - On this note, running a formatter on files will cause a lot of non-functional changes that have potential to conflict, and also makes it harder to see what actually changed in them. For instance, it's hard to see what changed in VMRouteFinder. (i can't really imagine what would need to?)
  - Not that I disagree with formatting changes, they just need to be applied quickly and ideally by themselves. I can review the changes currently in the PR though, so you don't need to change them back.
- The primary platform for the archive client is actually **WindowsDX**, since it performs way better and has less bugs. I'm not sure if newer versions of Monogame have fixed the biggest issues with the GL backend:
  - Resources were not properly ready by the first draw, so it appeared corrupt. This resulted in corrupt thumbnails and 3D lot facade generation.
  - Newer NVIDIA GPUs sometimes present an incomplete frame when loading assets (eg. the UI appears to flash)
  - AMD driver on Windows has very slow OGL compared to DX. I think their newer drivers tackle this, but I've experienced firsthand that their GL driver can be a bit broken on anything that isn't Minecraft.
- NativeAOT is an interesting target, but I'm a bit hesitant about removing Ninject. It's very tightly coupled to how the game passes around "global" state/services (within the kernel) and it's a huge surface area to replace. 
  - Back in the iOS port days I was able to keep Ninject using a combination of ugly manual calls to ensure functions could be reflected by Ninject, and I think a version of Ninject that worked without `System.Reflection.Emit` explicitly meant for targets like Xamarin.iOS. I would be surprised if they don't have a solution for NativeAOT, though maybe it's not maintained enough now to add one.
- When moving dependencies to nuget packages or newer versions, remember there can be a few tiny changes I made to patch bugs or improve performance.
  - For example, there was one change to Mp3Sharp to open files by name with more `FileShare` permissions, though I should really have just passed in the stream myself.
  - FSOMonogame has a few fixes, and some hacks for iOS that are no longer relevant. There was one big performance improvement to use a HashSet for resource dispose or something, which was big for FreeSO since it had a huge number of resources. These might be fixed upstream now.

TSOClient/tso.client/Rendering/City/CityVertex.cs
```diff
@@ -1,55 +1,55 @@
...
-            this.TextureWeight1 = new Vector4(terrain == TerrainType.Grass ? 1 : 0,
-                                              terrain == TerrainType.Snow ? 1 : 0,
-                                              terrain == TerrainType.Sand ? 1 : 0,
-                                              terrain == TerrainType.Rock ? 1 : 0);
-
-            this.TextureWeight2 = new Vector4(terrain == TerrainType.Water ? 1 : 0, 0, 0, 0);
-        }
-    }
-}
+﻿///*This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0.
```
riperiperi: I guess you've noticed since you commented them, but some files in here didn't exist in the project file for FreeSO since they're really old. Here are the files that are meant to be in the City folder - anything else can probably be safely deleted:

image

Other/libs/TargaImagePCL/TargaImagePCL.csproj
```diff
@@ -1,66 +1,22 @@
...
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'ServerRelease|AnyCPU'">
-    <OutputPath>bin\ServerRelease\</OutputPath>
-    <DefineConstants>TRACE</DefineConstants>
-    <Optimize>true</Optimize>
-    <DebugType>pdbonly</DebugType>
-    <PlatformTarget>AnyCPU</PlatformTarget>
-    <ErrorReport>prompt</ErrorReport>
-    <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
+    <PublishAot>true</PublishAot>
```
riperiperi, line 11: Does having this flag force an AOT publish? We do use publish to make the release versions of FreeSO (and azure builds, which obviously don't work with these changes). Maybe this should be false until AOT is a viable target.

TSOClient/FSO.UI/Debug/FSOExceptionDisplay.Designer.cs
```diff
@@ -74,7 +74,7 @@ private void InitializeComponent()
             // 
             // pictureBox1
             // 
-            this.pictureBox1.BackgroundImage = global::FSO.Client.Properties.Resources.ico256;
```
riperiperi, line 77: These files aren't part of the project anymore, so they can be safely removed.
image

TSOClient/tso.content/Content/DX/Fonts/mobile.xnb
riperiperi, line 1: Backend specific font files were removed recently. All the mgcb files were changed to output the fonts to the upper directory with the new slimmer format (`/outputDir:../Content`) since they weren't backend specific.

TSOClient/tso.content/Content/Effects/2DWorldBatch.xnb
riperiperi, line 1: I didn't even know these were still here - the game should access the effects from within the backend directories.

TSOClient/tso.content/ContentSrc/TSOClientContent.mgcb
```diff
@@ -10,136 +10,133 @@
...
+/processorParam:Range=4
+/build:Fonts/simdialogue.ini
+
+#begin Fonts/trebuchet.ini
+/importer:FieldFontImporter
+/processor:FieldFontProcessor
+/processorParam:ExternalPath=msdfgen.exe
+/processorParam:Resolution=32
+/processorParam:Range=4
+/build:Fonts/trebuchet.ini
```
riperiperi: These should have `/outputDir:../Content` now. Also applies to the other mgcb files.

TSOClient/tso.client/UI/Screens/LotScreenNew.cs
```diff
@@ -1,163 +1,163 @@
-﻿using System;
```
riperiperi, line 1: Looks like more files that aren't in the project. They can be safely deleted:
image

TSOClient/tso.client/Rendering/Lot/Components/CubeComponent.cs
```diff
@@ -1,82 +1,82 @@
-﻿using System;
```
riperiperi, line 1: The entire Rendering/Lot folder should be deleted. It's a precursor to FSO.World.

TSOClient/tso.client/Rendering/Sim/SimModelBinding.cs
```diff
@@ -1,30 +1,30 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using SimsLib.ThreeD;
-using Microsoft.Xna.Framework.Graphics;
-using TSOClient.Code.Data;
+﻿//using System;
```
riperiperi: The `Rendering/Sim` folder should also be deleted, this is a precursor to the Vitaboy projects.

TSOClient/tso.client/Network/LotPacketHandlers.cs
```diff
@@ -1,31 +1,31 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Runtime.Serialization.Formatters.Binary;
-using System.Text;
-using TSOClient.VM;
-using TSOClient.Lot;
+﻿//using System;
```
riperiperi: Some files in the Network folder are old and can be deleted. Here's what it should look like:
image

TSOClient/tso.client/MessagesCache.cs
```diff
@@ -1,48 +1,48 @@
-﻿/*This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0.
-If a copy of the MPL was not distributed with this file, You can obtain one at
-http://mozilla.org/MPL/2.0/.
+﻿///*This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0.
```
riperiperi, line 1: This file can be deleted.

TSOClient/FSO.Server/Servers/Api/Controllers/Admin/AdminShardOpController.cs
```diff
@@ -1,14 +1,8 @@
-﻿using FSO.Server.Database.DA;
-using FSO.Server.Protocol.Gluon.Model;
+﻿using FSO.Server.Common;
```
riperiperi: `FSO.Server/Servers/Api/` should no longer exist, which should make it possible to remove Nancy.

riperiperi (commented):
Looks great! The archive server seems to still work without changes, which is surprising given the horrors I had to pull to get the sqlite binary to copy properly on .NET framework. I guess on newer .NET, it just works.

Performance is notably better, even without any manual changes. Here's a comparison with a very high draw number of draw calls:

### Before:
image

### After
image

The weather's different, but that's not too important as the draw count should be similar, and the rain effect isn't very expensive.

Some initial findings after skimming the code and trying it out:

- Loading job lots in sandbox mode throws this error with the XmlReader: image
  - These XMLs are only ever included with or produced by the client so DtdProcessing can be set to true without much worry.
  
- The FSO.Server project needs to exclude the Admin folder, as node_modules can contain cs files for build scripts: image

- Mp3 playback needs to pass a readonly stream or it'll explode: image
  - Mp3 playback has an issue where it stutters for a long time at the start of playback. Not sure why that is. You can hear it happen when turning on any speakers on a lot.

- FSO.Server project needs to copy NLog.config to the output folder. (PreserveNewest) Somehow even after doing this, the logging doesn't work:
  - `no configuration section <common/logging> found - suppressing logging output`
  
- Randomly when joining a lot, the join process aborts and I can't enter any properties ever again. Maybe this is a race condition due to initialization moving a lot faster, but I can't check with NLog broken.
  
- The client process doesn't close properly sometimes when closing the game. I can probably debug this one myself after merge.

I noticed that scrolling rapidly across the city now causes some pretty big stutters while in a lot. Not really able to do CPU profiling due to the mystery connection hang but it could be fixed by this change I made to monogame ages ago:

https://github.com/riperiperi/FSOMonoGame/commit/1eaeeb10d7dcae2e3029d0b71ecffdf2545e494d#diff-d007e0b0bbbbf2eaf24fd40822b39cb4c74b12acb3e68d94c34ba3ffa4bc828cR117

... Though reintroducing it now would mean submoduling monogame again, or publishing our own variant on nuget.

Other/libs/TargaImagePCL/TargaImagePCL.csproj
SegerEnd: Since this is a class library, it isn’t AOT-compiled on its own. It only gets compiled AOT when the startup project that uses it it is published with AOT, so it currently does nothing

riperiperi: When I read this I assumed the flag was on every csproj, but I did notice it was missing from others.

riperiperi:
I'm getting some odd behaviour in Volcanic where a bunch of things seem to be the wrong size:
image
Here's what it looked like before:
image

I guess they've changed how a lot of these components render, and maybe they're also trying to adjust the sizing to better match modern Windows standards. When I open things in the designer and change them, it seems to update button sizes and positions:

image

The BHAV editor seems broken normally:
image

But after I resave the form from the designer, it fares a little better (still a lot bigger than it used to be):
image

I guess the runtime tries to resize things itself but ends up doing it worse than the designer. Not sure if they've documented this anywhere, or a way to work around it. It could just mean I need to go back and manually clean up all of the components to look the way they used to.

riperiperi:
I figured it out, it attempts to rescale based on the default font.
image

Adding this line to Program.cs fixes it at runtime:
image
`Application.SetDefaultFont(new Font(new FontFamily("Microsoft Sans Serif"), 8.25f));`

The designer still tries to change the size. I think adding an ApplicationDefaultFont in the csproj should influence the designer, though it still seemed broken when I tried:

`<ApplicationDefaultFont>Microsoft Sans Serif, 8.25pt, style=regular</ApplicationDefaultFont>`

TSOClient/tso.client/Rendering/City/CityVertex.cs
SegerEnd: I was already wondering why it didn’t want to compile with them, makes sense since they were outdated and using a different namespace also. I’ve deleted them completely now!

TSOClient/tso.content/Content/DX/Fonts/mobile.xnb
SegerEnd: Swapped the `.mgcb` files with your new ones (only updating the MSDF dll path for the net9.0 directory). Also placed back the font content path to the upper dir in `TSOGame`

TSOClient/tso.client/Rendering/Sim/SimModelBinding.cs
SegerEnd: Deleted :)

TSOClient/FSO.Server/Servers/Api/Controllers/Admin/AdminShardOpController.cs
SegerEnd: Removed them, Nancy NuGet dependency is gone

SegerEnd:
Hi! Thank you for the review and for taking the time to test it. The performance improvements are awesome to see, it really shows what improvements the newer version can do already ☺️.

I’m working through your notes and making improvements. A few updates from my side:

- **Archive server** - I currently have not a way to test in “archive mode” myself yet. If there’s a downloadable version, so I can help with the archive changes and make sure I’m getting the same behavior. As it currently it tries to download from localhost server `archive.ini` / `archive_test_2.zip` :)

- **WindowsDX** - I’ve made it the default again. It works the old way again using the MonogameLinker class, resolving the assembly from the platform dll, The `monogame.framework.dll` is also retrieved via NuGet and placed in the correct platform folder in [`FSO.Windows.csproj`](https://github.com/SegerEnd/FreeSO/blob/dotnet9/TSOClient/FSO.Windows/FSO.Windows.csproj) via a Target:
```xml
<Target Name="CopyMonoGameDLLs" AfterTargets="Build">
```
- **Format-on-save** - My Visual Studio was happily reformatting everything I saved. I’ve turned it off for now so future diffs won’t be noisy.

- **The VMRouteFinder change:** - I updated the `using` statement from `using TSO.Files.formats.iff.chunks;` to `using FSO.Files.Formats.IFF.Chunks;` so it would compile again. Bit hard to spot yes, with all the formatting happing.

- **Sandbox XMLs** - I’ve set DtdProcessing to Parse. The job lots now load again, so that one bug is fixed.

- **FSO.Server Admin folder** - I’ll excluded the Admin folder from the project to avoid pulling in the node_modules with the cs files.

- **Files removed** - Deleted all the old/unused files and directories you mentioned, so the project is cleaner now. I could maybe also further remove the libraries in _Other/_ that we don’t use anymore (e.g., NuGet replacements).

- **MP3 playback**  - I replaced the path with the FileStream,  and I am going to try and look into that startup stutter.

- **Volcanic IDE** - I added the `<ApplicationDefaultFont>`, I thought about maybe migrating to something like Eto.Forms for cross-platform support, but that would require a full rewrite since it doesn’t support the designer. That would be a lot of work, just updating the IDE sizes seems like a good way.

## next..

- **Simitone** - Probably Simitone also needs to be migrated to .NET 9.0 since it uses FSO as a submodule. It shouldn’t be too hard to migrate i think, so that’s likely my next step when this is merged into Archive.

riperiperi:
For Volcanic, I don't think `ApplicationDefaultFont` is enough, it still tries to resize the forms at runtime unless `Application.SetDefaultFont(new Font(new FontFamily("Microsoft Sans Serif"), 8.25f));` is in the FSO.IDE main function. `ApplicationDefaultFont` does change the font selected by the designer in Visual Studio, but the component placements still end up wrong.

I can't release the archive data right now as it hasn't been fully stripped of sensitive data (and it would mean essentially publishing a release with a lot unfinished). I guess it's possible to make an "empty" archive template, but that might take a while to do. I'll try and look at the NLog and lot server connection handshake issues myself, since I know the most about the process anyways it should be less trouble for me.

TSOClient/FSO.Server.Database/FSO.Server.Database.csproj
```diff
@@ -1,410 +1,39 @@
...
-    </ProjectReference>
+    <PackageReference Include="Dapper" Version="2.1.66" />
+    <PackageReference Include="MySql.Data" Version="9.4.0" />
+    <PackageReference Include="Ninject" Version="3.3.6" />
+    <PackageReference Include="System.Data.SqlClient" Version="4.9.0" />
+    <PackageReference Include="System.Data.SQLite.Core" Version="1.0.119" />
   </ItemGroup>
+
   <ItemGroup>
-    <Content Include="DatabaseScripts\changes\0003_max_roomates.sql">
```
riperiperi, line 314: These should be copied to the output directory so that the server db-init script works. Might change how this is done in future though since the archive _client_ doesn't really need it.

TSOClient/FSO.Server/FSO.Server.csproj
```diff
@@ -1,434 +1,67 @@
...
-      <Install>false</Install>
-    </BootstrapperPackage>
+    <Reference Include="Mario">
+      <HintPath>..\tso.world\Mario.dll</HintPath>
+    </Reference>
   </ItemGroup>
+
   <ItemGroup>
-    <None Include="NLog.config">
+    <None Update="config.json">
```
riperiperi: Does this file exist? `config.sample.json` does, but it doesn't seem to be copied to the output directory now...

TSOClient/tso.client/GameContent/ContentManager.cs
```diff
@@ -1,677 +1,677 @@
...
-using LogThis;
-using FSO.Client.Utils;
-using FSO.Client.UI.Framework;
-using FSO.Files.FAR3;
-using FSO.Client;
-
-namespace FSO.Client.GameContent
-{
-    public delegate void OnLoadingUpdatedDelegate(string LoadingText);
+﻿///*This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0.
```
riperiperi: This file can also be deleted.

TSOClient/tso.client/Rendering/City/CityCamera.cs
```diff
@@ -1,314 +1,314 @@
...
-            {
-                _AspectRatioMultiplier = value;
-                _DirtyProjection = true;
-            }
-        }
-
-        #endregion
-    }
-}
+﻿//using System;
```
riperiperi: This can be deleted.

TSOClient/tso.client/Regulators/LoginRegulator.cs
```diff
@@ -43,26 +40,35 @@ public LoginRegulator(AuthClient authClient, CityClient cityClient, IShardsDomai
...
         {
             switch (newState.Name)
             {
                 case "AuthLogin":
                     var loginData = (AuthRequest)data;
-                    var result = AuthClient.Authenticate(loginData);
+                    AuthResult result = null;
+                    try
+                    {
+                        result = await AuthClient.Authenticate(loginData); // Await the async method
```
riperiperi: Making this method async will mean that it now returns when it begins awaiting- before it's completed. I don't think anything's relying on this behaviour in this particular case but it seems pretty messy overriding a void function with an async void implementation.

TSOClient/FSO.UI/FSO.UI.csproj
```diff
@@ -1,210 +1,56 @@
...
+    <ProjectReference Include="..\tso.world\FSO.LotView.csproj" />
   </ItemGroup>
+
   <ItemGroup>
-    <Content Include="Dependencies\GOLDEngine.dll" />
-    <Content Include="Dependencies\Nuclex.Fonts.Content.TrueTypeImporter.dll" />
-    <Content Include="Dependencies\nunit.framework.dll" />
-    <Content Include="Dependencies\TargaImage.dll" />
-    <Content Include="discord-rpc.dll">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
riperiperi, line 199: I think this needs to be added, otherwise all of the discord rpc methods will fail. There's probably a better way to include this DLL now, but the solution at the time was just a bunch of DllImports in `FSO.UI.Model.DiscordRpc`.

TSOClient/FSOFacadeWorker/Program.cs
```diff
@@ -142,10 +137,10 @@ static void Main(string[] args)
         private static bool LoginSent;
         private static int Done;
 
-        public static void Login()
+        public static async Task LoginAsync()
         {
             api = new ApiClient(Config.Api_Url);
-            api.AdminLogin(Config.User, Config.Password, (result) =>
+            await api.AdminLoginAsync(Config.User, Config.Password, async (result) =>
```
riperiperi: If you discard the Task with `_ = await api.AdminLoginAsync...`, this method doesn't need to be async.

riperiperi (commented):
I figured out the issue I was having, it's an SQLite problem where it puts an int32 into the database instead of a uint32, then errors trying to cast it on retrieval. I've pushed a workaround on the archive branch, since it might actually happen there, too.

I don't think this should be enabled on any of our projects:
`<Nullable>enable</Nullable>`
It'll just cause warnings everywhere right now, and would need the whole project to be reworked to properly annotate everything.

About the AppDomain use in the server updater, I can't really remember why I added it in the first place. The solution I settled on later on was just alternating running the patcher and the game using a watchdog script, so it's probably fine.

For the `ApiClient` methods that you've made async, you should use `_ = ...` to discard the explicitly discard the tasks so that the warnings don't appear. Awaiting them doesn't do anything useful since the data comes back on the callback.

I haven't been able to test the normal server or update process, but it's likely those are broken on the archive branch anyways so it's better to fix them when I come back around to cleaning everything up.

Things are looking pretty good now. Whenever you want a final review set the PR as non-draft and I'll give it another look.

riperiperi:
Did some profiling and looks like MonoGame still has the O(n) dispose time that gets worse the more resources you have loaded (so if you dispose them all, it's O(n^2). Joining the town hall and scrolling rapidly across the map gets me big stutters.
image

The game tends to dispose assets in big chunks in both the city and lot views, but it being magnitudes slower isn't great even if I do distribute it over a few frames. I'll see if I can update FSOMonogame to the latest version, repatch the issue and set up nuget packages.

francot514:
Which Monogame is used, it allows the custom modifications?

SegerEnd:
> Which Monogame is used, it allows the custom modifications?

Currently, its using the official latest `3.8.4` MonoGame NuGet packages, I saw that @riperiperi upgraded the modifed FSOMonoGame to the latest MonoGame version so we can switch to that soon I think.

francot514:
Still youre using the Monogame linker classes or just package dependency

riperiperi:
It's using the Monogame linker so arguments can be passed to launch the game with a different backend.

TSOClient/FSO.Windows/Program.cs
```diff
@@ -32,6 +32,8 @@ public static void Main(string[] args)
                 var startProxy = new GameStartProxy();
                 startProxy.Start(UseDX);
             }
+
+            Environment.Exit(0);
```
riperiperi: Monogame is meant to call `TSOGame.OnExiting`, but for some reason it isn't doing that. Maybe they broke the event at some point.

Without the event, the `RCWorkerLoop` threads never terminate.

riperiperi: The archive client (and simitone) also have some code that prevents the game from being closed that isn’t working anymore. Simitone uses it to show the save confirmation dialog, and the archive client uses it to safely shut down the server.

I’ll have a look at why - it does seem to add the closing event to the form, but setting e.Cancel doesn’t stop it from closing. Maybe these issues are connected.

francot514: What about Environment.Exit(0); it is usable?

riperiperi:
I think I've solved the issue with the game not closing as well as properly waiting for the archive server to shut down with this change:

https://github.com/riperiperi/FreeSO/commit/d803ce7808012f0eb441514e0d35fe65f60f3203

Let me know if it fixes it for you. You can cherry-pick it into this branch if it works.

SegerEnd:
Yes, that change works for me, the game closes cleanly now on both IDE and Windows. I’ve cherry-picked it into this branch. Thanks!

TSOClient/FSO.IDE/Program.cs
```diff
@@ -68,6 +68,8 @@ public void InitVolcanic(string[] args)
         {
             Application.EnableVisualStyles();
             Application.SetCompatibleTextRenderingDefault(false);
+            Application.SetDefaultFont(new Font(new FontFamily("Segoe UI"), 8.25f));
+            Application.SetHighDpiMode(HighDpiMode.PerMonitorV2);
```
riperiperi, line 72: The HighDpi modes don't seem to work properly when you have multiple monitors with mismatching DPI:

image

This happens after dragging from my 1x monitor to my 1.5x one, then back.

riperiperi (approved):
Things are looking good for now, apart from some small issues:

- The Dispose performance issue. My fork has published development packages with the change, but GitHub's nuget store is pretty much useless for public projects since you need to login with a PAT token to read any public packages. On top of that being annoying to set up. Visual Studio asks me to re-enter the PAT token every time I start it. I've made a PR to upstream this over on the MonoGame repository, but I'm not sure when it'll be merged: https://github.com/MonoGame/MonoGame/pull/9027
  - This isn't really blocking anything, it just makes the game a bit more stuttery.
- FSOFacadeWorker appears to be trying to use DesktopGL monogame instead of WindowsDX. The custom graphics device code only works on the DX backend.
- There are some visual issues in the IDE from changing the font (it's generally wider than the old one)
- The HighDpi mode set in FSO.IDE doesn't work properly.

If you're good with everything so far, unmark it as draft and I'll merge.

TSOClient/FSO.IDE/Program.cs
SegerEnd: Ah, that’s not good! Guess the DPI setting should to be removed

PR #285, opened 2025-10-05
UIListBox keyboard navigation

SegerEnd:
This pull request enhances the `UIListBox` control by adding keyboard navigation.

* Added support for the `Up` and `Down` arrow keys to move the selection up and down.
* Added support for the `Enter` key to trigger the double-click event when an item is selected.
* Added wrap-around logic when navigating past the first or last item.
* Updates the scrollbar loaction while moving with the keyboard.
* Improves the **Sandbox Mode list** for smoother navigation <sub style="color:gray;">(also applies to the Archive character selection, i think).</sub>

> Initially, it found it unclear to double-click to start a lot, this makes that interaction more user-friendly.

let me know if this is something you want or not

### DIFF
--- a/TSOClient/FSO.UI/Controls/UIListBox.cs
+++ b/TSOClient/FSO.UI/Controls/UIListBox.cs
@@ -7,6 +7,7 @@ using Microsoft.Xna.Framework.Graphics;
 using FSO.Common.Rendering.Framework.IO;
 using FSO.Common.Rendering.Framework.Model;
 using FSO.Common.Utils;
+using Microsoft.Xna.Framework.Input;
 
 namespace FSO.Client.UI.Controls
 {
@@ -276,6 +277,15 @@ namespace FSO.Client.UI.Controls
 
             if (m_MouseOver)
             {
+                if (state.NewKeys.Contains(Keys.Up) && Items.Count > 0) 
+                    InternalSelect((m_SelectedRow - 1 + Items.Count) % Items.Count);
+
+                if (state.NewKeys.Contains(Keys.Down) && Items.Count > 0) 
+                    InternalSelect((m_SelectedRow + 1) % Items.Count);
+
+                if (SelectedItem != null && state.NewKeys.Contains(Keys.Enter)) 
+                    OnDoubleClick?.Invoke(this);
+                
                 var overRow = GetRowUnderMouse(state);
                 m_HoverRow = overRow;
             }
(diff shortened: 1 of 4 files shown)


TSOClient/FSO.UI/Controls/UIListBox.cs
```diff
@@ -276,6 +277,15 @@ public override void Update(UpdateState state)
 
             if (m_MouseOver)
             {
+                if (state.NewKeys.Contains(Keys.Up) && Items.Count > 0)
```
riperiperi, line 278: There's a focus context that might make a bit more sense to use than m_MouseOver. You would have to make the element take focus when the sandbox window appears (and more generally when any listbox is clicked). Focus is controlled by `state.InputManager`.

SegerEnd: I’ve added focus handling from `IFocusableUI` and set the focus to active in the `UIListBox` constructor, so it now applies to all list boxes by default. However, I can also change it so that focus is only set when the sandbox window appears, in case we don’t want every list box to receive focus upon creation.

riperiperi: I think it should only be the sandbox window. With this change the one in UILiveMode will take focus every time the mode switches to live, which might be odd (removes focus from the chat text box), and the admit list has two list boxes that it'll try to focus when it appears.

If I were redesigning the UI system from scratch I'd probably add some shared way for elements in dialogs to steal focus when they're shown. Maybe the show dialog methods can be changed to scan the element tree for `IFocusableUI` and pick one to focus, at least when the dialog is blocking.

riperiperi: Sort of related, buttons should have been made focusable so you can hit enter to dismiss dialogs, but a system was never put in place for focus navigation/highlight (and default focus) so it wasn't done.

francot514:
Good implementation of surrounds for sandbox, any plan to support neighborhood views for Simitone

SegerEnd:
> Good implementation of surrounds for sandbox, any plan to support neighborhood views for Simitone

Hi!, I didn’t implement the surround-lot functionality myself, I’m just calling the existing method from `VMLotTerrainRestoreTools` on enter when the option is enabled, so I don’t have to toggle the setting manually everytime.

francot514:
I have made a project for TS1 hybrid system if you want to check it

PR #289, opened 2026-02-05
Archive experiment

riperiperi:
I originally split this off due to using a custom version of Monogame and an experimental feature, but those things should be solved now.

There are a large number of changes in here to a lot of different things, the majority are around the lot connection lifecycle. Here are some highlights for other features:

- Github Actions build
- 3rd person direct control, improved direct control interaction with live mode UI / interactions
- Ambience rework for TSO and TS1
- HIT VM fixes for footstep sounds
- Archive server user and event management, dialog redesign
- RSA client/server automatic authorization (could still use a TLS connection)
- Huge performance improvements for loading and SimAntics heavy situations
- Fun archive specific features
- A lot of archive and game bug fixes + improvements (eg. TTS voice selection)

image


salvadorc17:
The archive experiment is just testing stuff no regular changes?

PR #292, opened 2026-02-09
Add server connection, Archive maker and Keyboard navigation

SegerEnd:
This PR adds support for connecting to a normal online server alongside Archive mode. You can now choose whether to connect to a server or an archive server from the same UI.

It also adds full keyboard navigation to the UI, improves cross-platform support for macOS and Linux.

## Server connection
Scherm­afbeelding 2026-02-09 om 00 54 46
  - Added Archive/Online radio toggle in the Join Server dialog so you can switch between archive mode and connecting to a server from the same UI

### Archive Maker

Added an Archive Maker script as temporary solution for creating playable archive builds. making it finally possible to test Archive mode before a stripped archive is publicly available. The script:

- Extracts the current state of a running Docker server
- Packages the server data into a .zip and serves it.

Not included in the PR anymore but placed separate, currently you can run instead:
```bash
curl -sLf https://raw.githubusercontent.com/SegerEnd/FreeSO/tools/docker/fso-archive.py | python3
```

### Docker Support (FreeSO Servers)
Added Docker support for running standard (online) FreeSO servers
Provides a consistent, reproducible environment for development and deployment
This makes it much easier to spin up, test, and deploy normal FreeSO servers without much setup.

### Bug fixes
  - **Missing SQL space in neighborhood query**: `"UPDATE fso_lots l"` was missing a trailing
  space before `SET`, causing SQL errors on neighborhood updates, e.g while purchasing a lot on a server.
  - **CityServer archive handshake null crash**: Added null check for `Config.Archive` in
  `CityServer.ArchiveHandshake()` so non-archive servers work.
  - **UIArchiveHostInformation null reference**: Added null check for `ShowIPButton?.Position` to
  prevent crash when the button doesn't exist. While pressing the "info" icon next to a lot name in Archive offline mode.
  - **SQLite library**: Replaced `System.Data.SQLite` with `Microsoft.Data.Sqlite`
  (10.0.2) across the entire server database layer for cross-platform compatibility. System.Data.SQLite is only Windows compatible. Some projects were already using `Microsoft.Data.Sqlite`
  - Fixed `GotoCAS()` always forcing archive mode (`controller.Archive = true` was hardcoded) breaking servers form functioning.

## Keyboard navigation
Adds keyboard navigation to the UI. You can now Tab/Shift+Tab between controls, use arrow keys and Enter/Space to interact, and press ESC to close dialogs.

### How Tab navigation works
- Tab cycles focus forward, Shift+Tab goes backward
- Focus order is based on screen position (top-to-bottom, left-to-right), or explicit `TabIndex` if set
- When a modal dialog is open, Tab is scoped to that dialog only
- Opening a modal dialog auto-focuses its first element
- When no UI element is focused and no dialogs are open, Tab passes through to gameplay (free cam)

### Controls with keyboard support
- **UIButton** - focusable, Enter to click/submit, hover state shown when focused
- **UIListBox** - arrow keys to navigate rows, Enter to select, mouse wheel to scroll
- **UITextEdit** - focusable, mouse wheel to scroll
- **UISlider** - arrow keys to adjust value, mouse wheel to scroll
- **UICombobox** - Enter/Space to open, arrow keys to navigate, Escape to close
- **UIRadioButton** - arrow keys to cycle within radio group
- **UIGridViewer** - arrow keys to navigate grid
- **UIClickableLabel** - focusable, Enter/Space to click
- **UIContextMenu** - arrow keys to navigate, Enter to select, Escape to close

 ## Native macOS and Linux support
  - FSO.Unix (name suggestions... ?) project for cross-platform support without using Mono, using modern .NET
  - macOS .app bundle creation with Info.plist and .icns icon
  - Native dialog support, instead of Winforms in FSO.Windows
  - Updated GameLocator paths for macOS to fix the bug that it would look in `~/Documents/Documents`
  - Tested on both MacOS and Linux Arch
  
| |
| :---: | 
FreeSO on Linux |
| **FSO on Linux Arch Hyprland** |

| | |
| :---: | :---: |
| FreeSO Archive on MacOS | Spectator mode |
| **FSO Archive: MacOS** | **FSO Online: Spectator Mode** |


TSOClient/tso.simantics/NetPlay/Model/Commands/VMNetAdjHollowSyncCmd.cs
```diff
@@ -55,7 +55,7 @@ public override bool Execute(VM vm)
                     }
                 }
 
-                if (VM.UseWorld && vm.Ready && !vm.FSOVAsyncLoading && !vm.Driver.RunningCatchup && changed && vm.Context.Blueprint.SubWorlds.Count != 0)
```
riperiperi, line 58: This is a new case triggered only by hollow sync that happens due to an event from a lot's LiveSurroundLotConnection. You can trigger this by joining two properties that are adjacent, making a change in build mode, then exiting build mode. The other property should display the change to the lot.

The check here prevents it from trying to update the surround lots while the graphics are loading for the first time - is there a reason it's being removed?

TSOClient/tso.client/UI/Screens/CoreGameScreen.cs
```diff
@@ -768,6 +768,12 @@ public void ClientStateChangeProcess(int state, float progress, UpdateState upda
                     break;
 
                 case 6: //done world load
+                    if (_pendingCacheClear)
+                    {
+                        TimedReferenceController.Clear();
```
riperiperi: This is not a good time for the cache clear - we want to minimize the stutter when the lot transition animation begins. It was also done twice due to how the CachableTexture2D destructor works.

TSOClient/FSO.Server.Database/DA/SqliteDAFactory.cs
```diff
@@ -16,8 +16,7 @@ public SqliteDAFactory(DatabaseConfiguration config)
             this.Config = config;
 
             // TODO: pass config connection string
-            // _pool = new SqliteConnectionPool("Data Source=fsoarchive.db;Version=3;UTF8Encoding=True");
-            Environment.SetEnvironmentVariable("SQLite_NoPlugins", "true", EnvironmentVariableTarget.Process);
+            // _pool = new SqliteConnectionPool("Data Source=fsoarchive.db");
```
riperiperi: The `SQLite_NoPlugins` option is important to stop it from scanning the archive data filesystem, which takes a very long time with the real archive data.

TSOClient/tso.client/UI/Screens/CoreGameScreen.cs
riperiperi: This is not a good time for the cache clear - we want to minimize the stutter when the lot transition animation begins. It was also done twice due to how the CachableTexture2D destructor works.

TSOClient/tso.simantics/NetPlay/Model/Commands/VMNetAdjHollowSyncCmd.cs
riperiperi, line 58: This is a new case triggered only by hollow sync that happens due to an event from a lot's LiveSurroundLotConnection. You can trigger this by joining two properties that are adjacent, making a change in build mode, then exiting build mode. The other property should display the change to the lot.

The check here prevents it from trying to update the surround lots while the graphics are loading for the first time - is there a reason it's being removed?

TSOClient/tso.world/World.cs
```diff
@@ -1189,7 +1189,7 @@ public bool Preload(GraphicsDevice gd)
 
             bool shouldReturn()
             {
-                if (watch.ElapsedMilliseconds > 16)
+                if (watch.ElapsedMilliseconds > 25)
```
riperiperi: This time value was 16ms to improve responsiveness during loading to be a bit closer to a 60fps target. Why has it been changed to 25?

TSOClient/tso.world/World.cs
```diff
@@ -1200,13 +1200,12 @@ bool shouldReturn()
...
             if (PreloadProgress == 0)
             {
-                var done = 0;
                 for (int i = PreloadObjProgress; i < Blueprint.Objects.Count; i++)
                 {
                     var obj = Blueprint.Objects[i];
                     obj.Preload(gd, State);
                     PreloadObjProgress++;
-                    if (done++ >= 6 && shouldReturn()) return false;
+                    if (shouldReturn()) return false;
```
riperiperi: This case prevents object preload from returning if it hasn't done much work at all, which means that on slower systems the game will prefer framedrops over decreasing the loading speed.

TSOClient/tso.simantics/Marshals/VMMarshal.cs
```diff
@@ -53,12 +53,7 @@ public void Deserialize(BinaryReader reader)
                 var length = reader.ReadInt32();
                 var cStream = new MemoryStream(reader.ReadBytes(length));
                 var zipStream = new GZipStream(cStream, CompressionMode.Decompress);
-                var decompStream = new MemoryStream();
-                zipStream.CopyTo(decompStream);
-                decompStream.Seek(0, SeekOrigin.Begin);
-                reader = new BinaryReader(decompStream);
-                cStream.Close();
-                zipStream.Close();
+                reader = new BinaryReader(zipStream);
```
riperiperi: This is done to fully decompress the marshal before reading individual fields out of it, which saves an unexpectedly high amount of time. It's written a bit unusually due to some odd behaviour with the GZipStream.

TSOClient/FSO.Server/ToolSqliteImport.cs
```diff
@@ -839,12 +839,6 @@ public int Run()
         {
             SetPragmas();
 
-            MigrateInventoryState();
```
riperiperi, line 842: Where has this gone? The SQLite version of the database requires that object state (saved when putting an object into a user's inventory) is inserted directly into a database column instead of in loose files.

Moving around the loose files takes a very long time for the full archive data, which is why I migrated it to the database.

TSOClient/FSO.Server/ServerConfiguration.cs
```diff
@@ -24,6 +24,9 @@ public class ServerConfiguration
         [JsonProperty("updateBranch")]
         public string UpdateBranch;
 
+        [JsonProperty("allOpenable")]
+        public bool AllOpenable;
```
riperiperi, line 28: I don't think this option is appropriate for a normal game server.

The option deliberately disables admission rules, and allows anyone to open a fully writable instance of any lot in the game without the presence of the owner. In TSO terms, it allows you to fill most lots in the game with trash piles and puddles, or open a lot just long enough for all of their flowers to die.

If something like this were available in the normal game server, it would have to:
- Open a read-only version of the lot that is instantly reopened as writable when the owner/roommates attempt to open it. (making sure nobody is fully kicked is particularly difficult)
  - While empty lots are currently opened as read-only, they would also need to be reopened when someone _buys_ the plot that people are standing on, then joins it.
- Respect ban rules in some way:
  - Prevent sims from entering any doors
  - Prevent sims from accessing any plugin data, such as signs (maybe outdoors is allowed if the property type isn't admit list or ban all)
  - Eject sims who are banned when the lot opens for real

TSOClient/FSO.Server/Embedded/ArchiveConfigBuilder.cs
```diff
@@ -26,7 +26,7 @@ public static ServerConfiguration Build(ArchiveConfiguration config)
                 Database = new Database.DatabaseConfiguration()
                 {
                     Engine = "sqlite",
-                    ConnectionString = $"Data Source={dbPath};Version=3;UTF8Encoding=True",
+                    ConnectionString = $"Data Source={dbPath}",
```
riperiperi, line 29: Where'd the other connection options go?

TSOClient/FSO.Server.Database/DA/SqliteDAFactory.cs
riperiperi: The `SQLite_NoPlugins` option is important to stop it from scanning the archive data filesystem, which takes a very long time with the real archive data.

TSOClient/FSO.Server.Api.Core/Controllers/InitialConnectController.cs
```diff
@@ -80,7 +80,6 @@ public IActionResult Get(string ticket, string version)
                 Response.Cookies.Append("fso", token.Token, new Microsoft.AspNetCore.Http.CookieOptions()
                 {
                     Expires = DateTimeOffset.Now.AddDays(1),
-                    Domain = Request.Host.Host,
```
riperiperi, line 83: Any reason for removing this? Couldn't browser based admin tools be interested in this?

TSOClient/tso.client/UI/Archive/UIArchiveHostInformation.cs
```diff
@@ -313,8 +313,7 @@ private void RecalculateSize()
 
             vbox.AutoSize();
 
-            var showIPBase = vbox.Position + ShowIPButton.Position;
-
+            var showIPBase = vbox.Position + (ShowIPButton?.Position ?? new Vector2(0, vbox.Size.Y));
```
riperiperi, line 316: This one's my mistake - the DiscordButton shouldn't actually be shown at all if the ShowIPButton isn't present.

TSOClient/tso.client/UI/Archive/UIArchiveLandingDialog.cs
```diff
@@ -36,7 +30,7 @@ public UIArchiveLandingDialog() : base(UIDialogStyle.Standard, true)
         {
             var ui = Content.Content.Get().CustomUI;
 
-            SetSize(496, 263);
+            SetSize(496, 293);
```
riperiperi: Why has this size changed? This dialog has a pretty strict design so it'll look weird if it's too tall.

TSOClient/tso.client/UI/Panels/UIUCP.cs
```diff
@@ -200,6 +202,9 @@ public UIUCP(UIScreen owner)
 
         private void InitArchive()
         {
+            var regulator = FSOFacade.Kernel.Get<CityConnectionRegulator>();
```
riperiperi: I don't like this way of checking this, but it's not consistent around the game yet since I hadn't planned switching back to the regular server yet.

TSOClient/FSO.Unix/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+                };
+                Process.Start(psi)?.WaitForExit();
+            }
+            else
+            {
+                try
+                {
+                    var psi = new ProcessStartInfo
+                    {
+                        FileName = "zenity",
```
riperiperi, line 74: Seems a bit silly to assume that this random application is in the path on linux. I'd prefer if it only did the console writeline.

TSOClient/FSO.Unix/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+        private static void ShowDialog(string text, string title)
+        {
+            if (text.Length > 1500) text = text.Substring(0, 1500) + "...";
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                var psi = new ProcessStartInfo
+                {
+                    FileName = "osascript",
+                    Arguments = $"-e \"display alert \\\"{Escape(title)}\\\" message \\\"{Escape(text)}\\\" giving up after 15\"",
```
riperiperi, line 63: This seems like an insane way to spawn a dialog, but I guess it only happens in emergencies, so whatever.

TSOClient/FSO.UI/UILayer.cs
```diff
@@ -325,10 +326,65 @@ public void Update(UpdateState state)
                 item.Update(state);
             }
 
+            HandleFocusNavigation(state, mainUI);
+
+            if (state.NewKeys.Contains(Keys.Escape) && Dialogs.Count > 0)
+                RemoveDialog(Dialogs[Dialogs.Count - 1]);
```
riperiperi: Not a good solution - a lot of dialogs need an answer or custom close handler, and some things are dialogs that you might not expect (such as ArchivePersonSelection)

TSOClient/FSO.UI/Utils/UIUtils.cs
```diff
@@ -167,6 +168,13 @@ private void Update(UpdateState state)
         {
             if (m_doDrag)
             {
+                if (state.MouseState.LeftButton == ButtonState.Released)
+                {
+                    m_doDrag = false;
+                    DragControl.RemoveUpdateHook(UpdateHook);
```
riperiperi: This is already done in `DragMouseEvents`? Is this working around something?

TSOClient/FSO.Server.Api.Core/FSO.Server.Api.Core.csproj
```diff
@@ -36,10 +36,11 @@
   <ItemGroup>
     <PackageReference Include="AWSSDK.Core" Version="3.3.103.48" />
     <PackageReference Include="AWSSDK.S3" Version="3.3.104.36" />
-    <PackageReference Include="Microsoft.Data.Sqlite.Core" Version="9.0.9" />
+    <PackageReference Include="Microsoft.Data.Sqlite.Core" Version="10.0.2" />
     <PackageReference Include="Octokit" Version="0.36.0" />
     <PackageReference Include="System.IO.FileSystem" Version="4.3.0" />
     <PackageReference Include="System.Security.Cryptography.Algorithms" Version="4.3.1" />
+    <PackageReference Include="System.IO.FileSystem.Primitives" Version="4.3.0" />
```
riperiperi: What's this used for?

riperiperi (commented):
Glad to see someone picking up on the support for the regular server, since I've been totally invested in getting the archive stuff moving. The improved focusable UI support will definitely help a lot with all the new configuration menus, and the Direct Control mode which requires focus switching a lot.

While the tool for building an archive from a database is useful (and it looks fantastic), the C# migration tools aren't complete yet, and are missing a lot of features (and documentation about what those features do and how to configure them) mostly relating to opening lot saves, something you can't exactly do from python. If this is added now, it will likely need to be updated to match future changes to the archive database structure or migration process. I will try to make it clear when this is fully locked-in.

Specifcally - there's an automated process that will delete sign data that's inaccessible, or on a property that is ban-all or admit-list (there's a human-in-the-loop version of this i'm using to manually make exceptions, too). There's another tool that scans properties for a _large difference_ between the newest and oldest backup, and prefers the oldest backup if it has considerably more objects from ex roommates (useful for counteracting the last save of a property being due to a move-out). I guess these tools will work on the database post-conversion, but it's not really set in stone yet.

The database shape might still change a little, too. There are some things on my TODO list that I don't think I put in the archive database generator:
- Anonymize fso votes
- Remove ban/admit list

**Some notes preventing a merge:**
- There are a number of changes to the VM and loading process that seem unrelated to anything here and will just mess with things that are either carefully tuned or partially complete (hollow lot reloading).
- There are some changes related to sqlite that might break compatibility with the archive data (specifically forcing UTF8 encoding). Is there a reason for them?
- AllOpenable option on the regular server. I explain more in one of the comments on the code, but it's designed for the more laid-back, freeform archive mode. Without modification, it's not really appropriate for an MMO server, as it's easy for users to be malicious with it.
- I think the behaviour where you can close dialogs with ESC can break the game in a lot of ways...

> Job lot desync: Moved LotActive.Set() and ActiveYet = true before the job lot
early-return path in LotContainer, so job lots properly signal readiness instead of hanging

I'm not really seeing this - I recently changed this to fix a desync caused by the new tighter loading process. Did it get lost in translation, or is it no longer an issue?

alexjyong:
+1 docker set up. This will make it easier for people to launch their own FreeSO instances.

TSOClient/FSO.Server.Api.Core/FSO.Server.Api.Core.csproj
SegerEnd: Sorry, it was a left over from my NET 9.0 to MacOS/iOS branch, doesn’t need it, compiles fine now without, removed it!

TSOClient/FSO.Unix/Program.cs
SegerEnd: Yes that’s true, I wanted to find a way to spawn dialogs, because we’re not on windows I can’t use winforms so this seemed like a simple alternative.

TSOClient/FSO.Unix/Program.cs
SegerEnd: It’s a pre installed system package on most Linux distro’s. That’s why I included it as fallback. But I’m willing to remove it and just always fall back on console :)

TSOClient/FSO.Server/Embedded/ArchiveConfigBuilder.cs
SegerEnd: Version=3 and UTF8 Encoding=True are System.Data.SQLite connection
options they're not supported by Microsoft.Data.Sqlite and would cause errors running it.

SegerEnd: I changed the dependency to Microsoft.Data.Sqlite because on System.Data.SQLite only Windows is supported causing MacOS and Linux to not play in Archive mode, getting a runtime exception

SegerEnd:
> While the tool for building an archive from a database is useful (and it looks fantastic), the C# migration tools aren't complete yet, and are missing a lot of features (and documentation about what those features do and how to configure them) mostly relating to opening lot saves, something you can't exactly do from python. If this is added now, it will likely need to be updated to match future changes to the archive database structure or migration process. I will try to make it clear when this is fully locked-in.

I agree, it was just a quick script for me to try and get a “blank” Archive server running for testing. It wasn’t meant to be a long-term solution. Python isn’t really my go-to :) just spin something quick up without external dependencies. I removed it from the PR. For ones like me that also find it useful for testing, currently you can use it for some time with this command directly from the CLI in the FreeSO/docker directory:

```bash
curl -sLf https://raw.githubusercontent.com/SegerEnd/FreeSO/tools/docker/fso-archive.py | python3
```

TSOClient/FSO.Server/ServerConfiguration.cs
SegerEnd: Yes, I see what problems it can cause on “public” servers enabeling this. If it’s going to be added, it must be well implemented. I’ll look into that. Sounds good to create, introduce a type of spectator mode.

TSOClient/tso.client/UI/Panels/UIUCP.cs
SegerEnd: Currently working on a finding a better alternative maybe in CoreGameScreen with something like?: `if (Mode == CityConnectionMode.ARCHIVE) view.ucp.InitArchive(); `

TSOClient/FSO.Unix/Program.cs
riperiperi: I had a closer look and it's probably fine, as long as it has that console fallback when the command isn't available.

TSOClient/FSO.Server/ToolSqliteImport.cs
riperiperi: I see what's happened here - there was an early return I put in because I only needed to run the inventory migration part, so the change to revert that does make sense.

SegerEnd:
I am currently working on a Spectator mode for online servers to replace the `"allOpenable"` flag in the server config. [https://github.com/SegerEnd/FreeSO/tree/spectator-mode](https://github.com/SegerEnd/FreeSO/tree/spectator-mode) _(Merging later in this PR branch)_

Its currently working good and in playable state, with both server-side and client-side restrictions in place, i'm going to do some code cleanup of my changes and fix the archive mode that is using currently opening all lots as a spectator _(its currently not looking at the archive configuration)_. Since I’m getting familiar with most parts of the overall architecture, I dont know if everything has been implemented at the appropriate places. I’d appreciate any feedback or suggestions for improvement.

Edit: When a admin/lot owner/roommate joins the lot, the lot smoothly transitions into a normal host lot and players can interact with  objects again. Without ejecting current spectators on the lot.

salvadorc17:
Whats spectator gameplay idea?  No sim selected or interactions done?

riperiperi:
Looks pretty promising, at a glance the changes seem to be in the right places. The transition to becoming a real lot might be one to be careful about - like someone's fish and plants could be killed by someone staying in the lot as a spectator for a long time, then saved for real once the owner joins. 

My idea for this was originally just fully closing and reopening the lot, but on further examination you could literally reload the lot without disconnecting anyone (the game already does this for a process called "self resync"), though it might be tricky.

- Pause synchronizing with clients, and save all avatar data to memory. Fully reload the lot save, then run NetSimJoinCmd for everyone to put them in the new lot for real (with their previous data from memory). Finally, resume synchronizing with clients and perform a full sync to reload all the clients. They'll barely know what hit them.

You might be able to go farther and try to place them in the positions they were originally at, and try to restore things like their direct control state. It gets more complicated if you want to keep interactions and go-here (VM stack frames and interactions), as deleted objects could reappear when reloading the save, and it will mean the IDs for those objects won't be available. It's an unlikely scenario so it's probably fine for the transition to be a little rough and ignore these things.

None of this is necessary for an initial implementation, though. I think leaving the "stayed on the lot for too long, then the owner showed up" case alone would be enough for a first version.

The full close and reopen might still be required for joining the purchased version of a lot, as "empty" lots in free roam have a fake ID, reduced terrain generation complexity and no attachment to the nfs/database. I kind of needed to think of this scenario for archive mode anyways.

salvadorc17:
Its similar to Ts1 house without family and buy / build but without allowing to do that?

riperiperi:
Are you happy with how things are right now (including spectator mode)? I can review again - I've been doing some local work that'll conflict, but I can stash it and merge this first if everything is good.

SegerEnd:
> Are you happy with how things are right now (including spectator mode)? I can review again - I've been doing some local work that'll conflict, but I can stash it and merge this first if everything is good.

Yes it’s ready. The only feature not fully implemented yet is saving with dead plants and fish as you mentioned when an owner joins the lot with spectators. otherwise, I’m happy with it.

salvadorc17:
Release version would be good to test the changes

TSOClient/FSO.Server.Database/SqliteCompat/SbyteHandler.cs
```diff
@@ -15,8 +15,7 @@ public class SbyteHandler : SqlMapper.TypeHandler<sbyte?>
                 return null;
             }
 
-            // For some reason, the sqlite connection sees the value as unsigned.
-            return (sbyte)Convert.ToByte(value, CultureInfo.InvariantCulture);
+            return (sbyte)Convert.ToInt64(value, CultureInfo.InvariantCulture);
```
riperiperi, line 18: Were you encountering an issue with this? I guess maybe it will error for a negative byte, but I never saw this happen and don't know what column could ever trigger it.

TSOClient/FSO.UI/Controls/UIListBox.cs
```diff
@@ -310,6 +325,7 @@ private void OnMouseEvent(UIMouseEventType type, UpdateState update)
             {
                 case UIMouseEventType.MouseOver:
                     m_MouseOver = true;
+                    update.InputManager.SetFocus(this);
```
riperiperi: This will steal focus from anything else by accident a lot. If you want the mouse wheel to work when hovering, just allow that when `m_MouseOver` is true. The arrow keys should probably still follow focus rules.

TSOClient/tso.common/Utils/CurLoader.cs
```diff
@@ -77,7 +77,7 @@ public static Tuple<Texture2D, Point> LoadCursor(GraphicsDevice gd, Stream strea
                 outIO.Write(new char[] { 'B', 'M' });
                 outIO.Write(size + 14); //size, plus header
                 outIO.Write(0);
-                outIO.Write(14);
+                outIO.Write(62);
```
riperiperi: Why is this 62 now? This should be the offset to the pixel data, which appears right after the header. (14 bytes)

TSOClient/tso.simantics/NetPlay/Model/Commands/VMNetDirectControlToggleCmd.cs
```diff
@@ -36,6 +36,7 @@ public override bool Execute(VM vm, VMAvatar avatar)
 
         public override bool Verify(VM vm, VMAvatar caller)
         {
+            if (IsSpectator(caller)) return false;
```
riperiperi: This should be allowed. Though you should make sure that in the VMDirectControlFrame that it can't start interacting with any portals.

TSOClient/tso.simantics/Engine/Routing/VMRectRouter.cs
```diff
@@ -17,7 +17,7 @@ public VMRectRouter(VMObstacleSet map)
 
         public LinkedList<VMWalkableRect> Route(Point from, Point to, int startCardinal)
         {
-            var openSet = new List<VMWalkableRect>();
+            var openSet = new PriorityQueue<VMWalkableRect, int>();
```
riperiperi, line 20: Not a relevant change for this PR, but it's definitely an improvement.

TSOClient/FSO.Server/Servers/Lot/Domain/LotContainer.cs
```diff
@@ -102,7 +102,26 @@ public class LotContainer
         private LiveSurroundLotConnection SurroundConnection;
         private HashSet<uint> FreeRoamLeaving = [];
 
-        private bool AllowGuestOpening => Config.Archive != null && Config.Archive.Flags.HasFlag(FSO.Common.ArchiveConfigFlags.AllOpenable);
+        private bool AllowGuestOpening => Config.AllOpenable || IsArchiveServer;
+        private bool IsArchiveServer => Config.Archive?.Flags.HasFlag(FSO.Common.ArchiveConfigFlags.AllOpenable) ?? false;
```
riperiperi: Archive servers can have this unset, so it's weird to call this variable `IsArchiveServer`. Should probably be `ArchiveFreeRoam` or something.

TSOClient/FSO.Server/Servers/Lot/Domain/LotContainer.cs
```diff
@@ -1332,6 +1365,41 @@ public void BlockOnLotThread(Action action)
...
+            IsSpectatorMode = true;
+
+            // Set spectator flag on all remaining visitor avatars
+            foreach (VMAvatar ava in Lot.Context.ObjectQueries.Avatars)
+            {
+                if (ava.AvatarState.Permissions > VMTSOAvatarPermissions.Visitor) continue;
+                ((VMTSOAvatarState)ava.TSOState).Flags |= VMTSOAvatarFlags.Spectator;
+            }
+
+            // Self-resync reloads the VM and SyncAllClients pushes the updated state to all clients
```
riperiperi: Self-resync reloads the server VM, which isn't necessary. It's a mechanism to prevent the server from getting into a state that's impossible for clients to sync with. (eg stuck collision... mostly bugs)

I mentioned it earlier as being similar to what we want for the transition _from_ spectator mode, where we reload the lot in-place (which is more like ResetVM, but we want to keep the avatars)

Sending sync to all clients should be enough to update this state without a specialized command, so long as the save is made right now and it doesn't use a cached one.

TSOClient/FSO.Server/Servers/Lot/Domain/LotContainer.cs
```diff
@@ -1230,18 +1255,21 @@ public void Run()
                         TickFreeRoam();
                     }
 
-                    var beingKilled = preTickAvatars.Where(x => x.KillTimeout == 1);
-                    if (beingKilled.Count() > 0)
+                    if (!IsSpectatorMode)
```
riperiperi: Avatar data should be allowed to save. Would be weird if it dropped relationship changes just because the lot is transient.

TSOClient/FSO.Server/Servers/Lot/Domain/LotContainer.cs
```diff
@@ -1217,7 +1242,7 @@ public void Run()
 
                     if (--LotSaveTicker <= 0)
                     {
-                        SaveRing();
+                        if (!IsSpectatorMode) SaveRing();
                         LotSaveTicker = LOT_SAVE_PERIOD;
 
                         Host.UpdateActiveVisitRecords();
```
riperiperi, line 1247: This probably shouldn't count in spectator mode (the roomies shouldn't earn visitor hours when they aren't hosting). `RecordStartVisit` should also probably not fire until the avatar becomes non-spectator, and the lotvisit cleanup at the bottom of `ReleaseAvatarClaim` should happen if the avatar becomes spectator (and set the visitId to null).

TSOClient/FSO.Server/Servers/Lot/Domain/LotContainer.cs
```diff
@@ -1276,6 +1304,11 @@ public void Run()
                         while (lotActions.Count > 0) lotActions.Dequeue()();
                     }
 
+                    if (ShouldTransitionToSpectator)
```
riperiperi: There's quite a bit hiding in this call, and it'll check every tick. I think instead it should check `(!AllowGuestOpening || IsArchiveServer || IsSpectatorMode)` and transition when it would normally try to close the lot due to `noRoomies` (after a timeout, not instantly).

Honestly, I'm not sure the transition to spectator can work without a bunch of additional work. Consider the following:

- The host leaves, and 20 players are on the lot using skill objects indoors.
- If we leave them there and enter spectator mode, they're trapped inside, and they're in an interaction when they shouldn't be.
- We'd need to forcibly cancel their interaction then teleport them outside.
  - The server also forcibly cancels interactions to remove people from lots, or remove them from saves during ResetVM.

It might be best just to boot everyone in this case for now (as usual) and force them to rejoin, but honestly it's a bit of a similar case to reloading when the lot leaves spectator mode, so solving that could be a path to solving both.

TSOClient/FSO.Server/Servers/City/Domain/LotAllocations.cs
```diff
@@ -256,6 +260,29 @@ private Task<TryFindLotResult> TryFind(uint lotId, uint avatarId, bool openIfClo
...
+                                    {
+                                        isAdmin = db.Avatars.GetModerationLevel(avatarId) > 0;
+
+                                        if (lot.lot_id != 0)
+                                        {
+                                            var roomies = db.Roommates.GetLotRoommates(lot.lot_id);
+                                            isRoommate = roomies.Any(r => r.is_pending == 0 && r.avatar_id == avatarId);
+
+                                            // Spectators still respect ban rules
+                                            if (!isRoommate && !isAdmin
```
riperiperi, line 274: My view on this is that the ban rules aren't too important for spectators if the following things are present:

- Players can't interact with objects on the property (done)
- When a roomie joins, the players that are not on the admit list or are on the ban list should be instantly ejected (not done)
- The lot state is reloaded so the player can't have done any damage by waiting for a long time. (not done)

It's probably alright to leave it respecting these rules for now, though it will make the free roam experience a bit painful if they run into a ban_all property while running around, since it'll just suddenly boot them to map screen.

Right now though, this doesn't respect "admit list", which should prevent people from joining a lot unless they are specifically allowed. You can break into one of these properties as things are by hanging around until the owner arrives.

riperiperi (commented):
Left some comments. Looks good, but I've got a few comments on the rules for spectator mode. We can still do a followup that implements the lot reload separate from this PR.

TSOClient/FSO.Server.Database/SqliteCompat/SbyteHandler.cs
SegerEnd: Yes I was, I could not buy a new empty lot as a new sim in Archive mode.

TSOClient/tso.common/Utils/CurLoader.cs
SegerEnd: I got en error at 14 bytes on both MacOS and Linux, It’s for Cursor support on Linux and MacOS, on windows it still works too. I don’t know all the correct offsets, but this seems to work correctly?

I found this online while searching this issue:

- 14 (file header)
- \+ 40 (BITMAPINFOHEADER)
- \+ 8 (palette data)
- = 62

I have now improved it by calculating the offset in the method instead of the magic number.

SegerEnd:
I’ve now added the functionality where when the owner or roommate joins a lot the lot gets reloaded from the previous save. And additionally, users inside a room are automatically returned to the mailbox outside.

SegerEnd:
Hi! Is everything okay with the current state? Or do I still need to do some changes for approval?

riperiperi:
I haven't had much time to work on the archive stuff, so I haven't been able to get around to doing a review. I'll try have another look ASAP, sorry!

TSOClient/tso.simantics/NetPlay/Model/Commands/VMNetInteractionCmd.cs
```diff
@@ -29,8 +29,13 @@ public override bool Execute(VM vm, VMAvatar caller)
...
         }
 
+        private static readonly uint PAYPHONE_GUID = 0x313D2F9A;
+        private static readonly uint NHOOD_PAYPHONE_GUID = 0x303CD603;
+
         public override bool Verify(VM vm, VMAvatar caller)
         {
+            if (IsSpectator(caller) && vm.GetObjectById(CalleeID) is VMGameObject obj
+                && obj.Object.OBJ.GUID != PAYPHONE_GUID && obj.Object.OBJ.GUID != NHOOD_PAYPHONE_GUID) return false;
             if (caller == null && FromNet) return false;
```
riperiperi: This check should happen first, as it eliminates invalid commands. It won't make any functional difference though, so it's fine for now.

TSOClient/FSO.Server/Servers/Lot/Domain/LotContainer.cs
```diff
@@ -1176,20 +1183,32 @@ public void Run()
...
                         }
 
                         if (--TimeToShutdown < TICKRATE * 10)
                         {
-                            //no roommates are here, so all visitors must be kicked out.
-                            if (preTickAvatars.Count > 0)
+                            if (AllowGuestOpening && !ArchiveFreeRoam)
                             {
-                                Host.Broadcast(null, new FSOVMProtocolMessage(true, "21", "22"));
+                                TransitionToSpectatorMode();
```
riperiperi, line 1203: Will this force community or admit mode 5/6 lots to transition to spectator? They're explicitly fully openable by guests, so shouldn't even try to shut down unless nobody is online. That's why the check overwrote `noRoomies`.

Also, not sure if this will cause unusual behaviour with archive free roam, where it will always run the shutdown timer and as soon as the last person leaves it will instantly close.

TSOClient/FSO.Server/Servers/Lot/Domain/LotContainer.cs
```diff
@@ -1332,6 +1351,66 @@ public void BlockOnLotThread(Action action)
...
+            {
+                try
+                {
+                    var path = Path.Combine(Config.SimNFS, $"Lots/{LotPersist.lot_id:x8}/state_{LotPersist.ring_backup_num}.fsov");
+                    using var file = new BinaryReader(File.OpenRead(path));
+                    var marshal = new VMMarshal();
+                    marshal.Deserialize(file);
+
+                    Lot.Load(marshal);
+                    ResyncTime();
```
riperiperi: I think it needs to do more than this. When loading a save, there are a number of things that might need to be corrected:

- might need to send back objects from avatars who aren't roomies
- move_flags mean the terrain needs to be regenerated, or the lot might even be entirely reset
  - Note that this means that lots opened in spectator mode shouldn't _consume_ the move_flags and set them to 0. I've kind of taken a different approach with the archive flags in my new branch, but it wasn't backported to the normal flags.
- Everything in Lot.TSOState might be out of date if the lot was renamed, moved etc.
- Tuning vars from server
- Entity reset; the save might have happened with avatars still on the lot, so they have to be removed (may need to simulate a bunch of ticks to safely cancel their interactions, this happens in `AttemptLoadRing`, `CleanLot`) and object state might have to be patched.
- Various lot/object state updates

For this reason, it's probably best that almost all of `ResetVM` is run again. There are some things that should be skipped:

- Creating a new VM and server driver. That means everything up to and including `Lot.Init();`. Maybe this could be moved to a separate function or made optional with an argument like `reuseVM = true`.
- Reloading the HollowLots, that would just be wasted time.

This would also mean there's no need to duplicate the ring loading behavior here.

TSOClient/FSO.Server/Servers/Lot/Domain/LotContainer.cs
```diff
@@ -1332,6 +1351,66 @@ public void BlockOnLotThread(Action action)
...
+            {
+                if (ava.KillTimeout != -1 || ava.AvatarState.Permissions > VMTSOAvatarPermissions.Visitor) continue;
+
+                foreach (var action in new List<VMQueuedAction>(ava.Thread.Queue))
+                    ava.Thread.CancelAction(action.UID);
+
+                if (mailbox != null && ava.Position != LotTilePos.OUT_OF_WORLD
+                    && !Lot.Context.RoomInfo[Lot.Context.GetRoomAt(ava.Position)].Room.IsOutside)
+                {
+                    SimAntics.Primitives.VMFindLocationFor.FindLocationFor(ava, mailbox, Lot.Context, VMPlaceRequestFlags.UserPlacement);
```
riperiperi: This is unsafe - interaction cancellation can happen over a few seconds, and the interaction script may assume that they're in a slot or specific position. For example, this could instantly eject a sim from a shower while they're still in the showering state, then the script will continue and try to place them outside the shower, indoors (or crash because it expected the sim to be contained).

The safest way to do this right now is via `CleanLot()` then `Lot.Reset()`, which will remove all sims from the lot, then you can re-add them to the VM similar to how you're doing it when transitioning from spectator mode. This does mean that sims outside don't retain their interactions or positions, but it avoids any unsafe scenarios. They can't interact with objects outside in spectator mode anyways, so not too much lost for now.

riperiperi (commented):
Looking better, some changes to how the reload is done and I think everything should be covered. I'll rebase my recent changes on top after this PR is merged, no need to worry about merge conflicts there.

Apologies again for the long review delay, I've been busy with other things and it takes a while to get into the proper headspace to review the lot container code (because it's a little bit of a mess of interconnecting checks and processes).

TSOClient/FSO.Server/Servers/Lot/Domain/LotContainer.cs
riperiperi: You might need a way to tell the server driver not to eject players when their sims are deleted by `CleanLot()`.

riperiperi: Deleted sim disconnection happens in `HandleClients()`, it'll run a few times when running CleanLot or any other function that calls `Lot.Tick`. This also processes client commands and inactivity timer, which won't be too useful when trying to reset the lot ASAP. Might be best to add a boolean that disables running HandleClients at all when it's transitioning.

riperiperi:
Is this ready for review? I notice that a few things changed but not sure if you're entirely done yet.

SegerEnd:
Yes, I believe it is complete. I have not been working on it recently, but I was having trouble getting some of your recent advice to work without causing the player to return to the city screen.

riperiperi (approved):
Looks great, thanks for all the work on this! I do have some things I'd like to fix and test out, but it's better that it gets merged with all the work I've been doing first to avoid any more merge conflicts.

Took a bit longer than I planned to review the PR since I wanted to be done with the feature I was working on in the `archive-experiment-temp` branch first, but it was a lot more work than I expected...

PR #299, opened 2026-05-17
Add eyedropper to FreeSO

SegerEnd:
This pull request brings some two new build-mode tools, The Eydropper and Sledgehammer.

**Key architectural updates:**

* I've added the `IWallHoverTool` interface and the `ArchitectureHit` to give tools a way to represent and interact with things like walls, floors, and objects that you hover over.

**New build-mode tools:**

* I’ve added the `UIEyedropper` and `UISledgehammer` classes. This lets you pick up, paste, or delete elements and objects, respectively. Both tools use the new picking logic.

**Refactoring and tool unification:**

* I’ve refactored `UIWallPainter` to include `IWallHoverTool` and made use of the new picking logic through `UIArchitectureTools.Pick`. So we can now paint directly on walls in 3D mode.

Scherm­afbeelding 2026-05-17 om 15 40 55

One thing I’m not sure about are the icons, which need to be improved,

And the usefulness of the Sledgehammer otherwise it can be removed.


SegerEnd:
I’ll leave it here for now until I clean it up and fix things, probably removing the sledgehammer.

riperiperi:
Let me know if/when you want a review. I think at a glance the only thing that needs improving is visibility on what the tool is hovering, but that's not present for other systems where it should anyways (objects are meant to highlight on mouse hover in live mode, but i chose not to add this) so could be a lot of work.

